PR #10, opened 2019-04-24
now, when exporting a group to csv, it is possible to include avatar …

DrumsnChocolate:
this pull requests adds the possibility to request avatar urls when exporting groups.
needs: https://github.com/csvalpha/amber-api/pull/21

### DIFF
--- a/app/controllers/groups/export.js
+++ b/app/controllers/groups/export.js
@@ -90,7 +90,7 @@ export default Controller.extend(FileSaverMixin, {
       label: 'Noodnummer'
     }, {
       value: 'avatar_url',
-      label: 'profielfoto url'
+      label: 'Profielfoto url'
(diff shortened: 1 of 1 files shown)


app/controllers/groups/export.js
```diff
@@ -88,6 +88,9 @@ export default Controller.extend(FileSaverMixin, {
     }, {
       value: 'emergency_number',
       label: 'Noodnummer'
+    }, {
+      value: 'avatar_url',
+      label: 'Avatar URL'
```
Matthijsy: Can you change the label to profielfoto url? We always call it profielfoto to the user

DrumsnChocolate: oh ja klinkt wel logisch

PR #16, opened 2019-05-05
Revert DIES theme

cmitz:
### Summary 
Revert Dies Theme as was added in https://github.com/csvalpha/alpha-banana-ui/pull/1376


Matthijsy:
Maybe also remove the navbar unauthenticated class from navbar since this was added as well in the PR

PR #17, opened 2019-05-05
User page layout fixes

cmitz:
### Summary
Update the styling on the User pages, fixing multiple errors and glitches. Also tested on mobile.

image


cmitz:
> Looks good! Maybe rebase since the Dies revert is also part of this PR now

I really don't like doing that, even though that commit was cherry-picked and merged in a seperate PR. It's mainly because I'd have to force-push the rebased branch which is dangerous.

PR #31, opened 2019-06-23
Improve Oauth

Matthijsy:
### Summary
This PR makes the UI responsibe for the oauth flow. See the API pr for more information.

I am not glad with the hacky implementation of `clientName` but do not know how to improve this. We need to load this endpoint automatically when this route is loaded, but cannot use the model since the clientName data needs to be available within the controller. It can proabably better but I do not know how. Please give feedback.

Needs: https://github.com/csvalpha/amber-api/pull/52

### DIFF
--- a/app/controllers/oauth/authorize.js
+++ b/app/controllers/oauth/authorize.js
@@ -1,38 +1,51 @@
 import Controller from '@ember/controller';
-import { inject as service } from '@ember/service';
+import {inject as service} from '@ember/service';
+import {computed} from '@ember/object';
 
 export default Controller.extend({
-  ajax: service(),
+  fetch: service(),
 
   /* eslint-disable camelcase */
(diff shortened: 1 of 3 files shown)


app/controllers/oauth/authorize.js
```diff
@@ -1,38 +1,51 @@
...
   /* eslint-enable camelcase */
 
+  clientName: computed('client_id', function() {
+    this.get('fetch').fetch(`/oauth/authorize?client_id=${this.get('client_id')}&redirect_uri=${this.get('redirect_uri')}&response_type=${this.get('response_type')}&state=${this.get('state')}`
+    ).then(response => {
+      response.json().then(json => {
+        if (json.status === 'redirect') {
+          location.href = json.redirect_uri;
+        } else {
+          this.set('clientName', json.client_name);
```
cmitz: Isn't it possible to just return clientName, instead of setting it?

PR #38, opened 2019-09-13
Update node to 12.10.0

cmitz:
### Summary
Update Node version


cmitz:
Yes we should

cmitz:
These are the only mentiones of Node versions in the whole project. Dockerfile uses a ember-specific-image by some dude.

Issue #39, opened 2019-09-16
YouTube regex breaks info pages with some links or images


wilco375:
### How to reproduce
1. Create a new info page or edit an existing one
2. Enter the following content and save the page: 
`[hier](https://forms.gle/PQRjWdhdg6fWDxQd9)!`
or 
<details><summary>
Expand
</summary>
<p>

```
## XXXVIIIe Senaat (2017-2018)

![](https://i.postimg.cc/XvMKQ1KV/ef141747-a446-4057-87c4-1cac337da355.jpg)

Bente Meijerink - Praeses  
Matthijs Vos - Abactis  
Mirjam Bos - Quaestor  
Nick van den Berg - Assessor   

## XXXVIIe Senaat (2016-2017)

![](https://i.postimg.cc/ZRvPCYjP/687474703a2f2f776573702e736e742e757477656e74652e6e6c2f7e616c7068.jpg)

Bart Treurniet - Praeses  
Beate Blom - Abactis  
Eeuwe Krikke - Quaestor  
Jan-Willem Nijhuis - Assessor Internorum  
Martijn Mulder - Assessor Externorum  

## XXXVIe Senaat (2015-2016)

![](https://i.imgur.com/wnOtqRY.jpg)

Dennis Hans - Praeses  
Len Spek - Abactis  
Rutger Frieswijk - Quaestor  
Ruben Smit - Assessor Internorum  
Paulien Engelbert - Assessor Externorum  

## XXXVe Senaat (2014-2015)

![](https://i.postimg.cc/tTGJNXpN/687474703a2f2f776573702e736e742e757477656e74652e6e6c2f7e616c7068.jpg)

Adriaan Pleijsier - Praeses  
Thijs de Wilde - Abactis   
Mike van der Wekken - Quaestor  
Edwin van Boven - Assessor  

## XXXIVe Algemeen Bestuur (2013-2014)

![](https://i.postimg.cc/xdMTCWMJ/687474703a2f2f776573702e736e742e757477656e74652e6e6c2f7e616c7068.jpg)

Joanne Lijbers - Praeses  
Denise Boele - Abactis  
Gunnar Slaa - Quaestor  
Marijn Romeijn - Assessor Internorum  
Mark de Vries - Assessor Externorum  

## XXXIIIe Algemeen Bestuur (2012-2013)

[...]

PR #41, opened 2019-09-19
Fix youtube regex

wilco375:
See issue #39

### Summary
This PR fixes the problem where the youtube regex would match urls that weren't youtube urls by making sure that the url is from the domain youtube(.com) or youtu(.be).

### DIFF
--- a/app/initializers/register-showdown-extensions.js
+++ b/app/initializers/register-showdown-extensions.js
@@ -2,12 +2,13 @@ import showdown from 'showdown';
 
 export function initialize() {
   showdown.extension('youtubeEmbed', () => {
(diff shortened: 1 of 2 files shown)


app/initializers/register-showdown-extensions.js
```diff
@@ -2,12 +2,13 @@ import showdown from 'showdown';
 
 export function initialize() {
   showdown.extension('youtubeEmbed', () => {
-    const youtubeRegex = /.*(\$\(youtu.be\/|v\/|e\/|u\/\w+\/|embed\/|v=)([\w-]+)\).*/g;
```
wilco375: The `youtu.be` part can be removed here, because in the new regex the `youtu.` domain is matched and the `e\/` matches the end of the `.be` domain

Matthijsy: Could you maybe extract the regex and include some tests? To prevent this kind of regression

PR #47, opened 2019-10-27
Remove FA pro

Matthijsy:
### Summary
The idea of this PR was to update all packages.. however that breaks everything so I will wait with that until we upgraded ember to the most recent version. This PR now only removes the pro variant of fontawesome, since we do not use it.


cmitz:
We've got a lot of places where the icon name is generated somewhere other than the handlebars template. For example the MD-editor. Please check if those icons are working as well...

Matthijsy:
I've tried to check it (visual) and think I've got everything. However, it is hard to catch the 'icon not found' logs in the console since this is filled with deprecation warnings so it could be that I missed one.

PR #56, opened 2019-11-06
hotfix/relationship-loading

Matthijsy:
### Summary
Currently the article-comments are not shown. This was because we are dropping link data (since this can be broken due to namespacing issues). However, the relationships was also not present in the 'data' field. This means that the relationship was gone. By only dropping the linkage data when the data field is present we force Ember to use the data field when present (we can use this when the links are wrong) and use the links as fallback. This way the article comments will work again and the models with wrong links (like poll) wil also work


cmitz:
Do the forum and activity pages work?

cmitz:
Looks like it works on staging 👍

PR #60, opened 2019-11-12
Hotfix activity link

Matthijsy:
### Summary
The 'closing activity' tool is broken. It loads the data and calls a 'pushPayload'. This is not serialized by `normalizeReponse` but it is by `normalize`, so I changed the logic to that method. 

The 'closing indicator' still shows wrong values sometimes but I cannot find why. At least the link is working again

### DIFF
--- a/app/serializers/application.js
+++ b/app/serializers/application.js
@@ -29,7 +29,7 @@ export default JSONAPISerializer.extend({
       });
     }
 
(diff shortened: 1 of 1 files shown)


cmitz:
Ziet eruit alsof dit altijd al de beter oplossing was 😅

Issue #22, opened 2019-05-10
Reduce bundle size


cmitz:
We need to reduce our bundle size. It's now a whopping 5 MB.

Some things we can do:
- Replace EmojiOne with unicode-only or another, much smaller, library
- Slim down bundled FontAwesome package ([maybe whitelist like this](https://github.com/csvalpha/amber-ui/blob/fix/fontawesome-md-toolbar/config/environment.js#L74))
- Remove jQuery
- Remove MomentJS in favour of a smaller library

There's probably a lot more, but that'd need some investigation using this: https://codingitwrong.com/2018/09/02/ember-bundle-size.html

Matthijsy:
Also consider something for showdown.js, this is even bigger than momentjs

PR #79, opened 2020-01-03
Specify used fa-icons

Matthijsy:
### Summary
Related to #22 
This PR explicitly lists all the fa-icons into the environment. This makes that our bundle only includes those icons. This reduces our bundle vendor.js size from 658 KB to 445 KB, saving us 200 KB of bundle size (around 30%).


Matthijsy:
Great catch!

PR #81, opened 2020-01-03
Remove unused images

Matthijsy:
### Summary
Related #22 
I didn't find any references to those images so I think they can be removed from the code base


wilco375:
Why do we even have that Kylo Ren image :thinking: 
Can't find any references either, so looks fine to me.

Matthijsy:
Probably used as test image during early development or something like that.

PR #85, opened 2020-01-09
Fix date example format, add start_study column

wilco375:
### Summary
The date was formatted inconsistenly, this is now fixed.
The start_study column was missing, which I added.


wilco375:
Yes, I tested uploading this file and it interprets it correctly.

PR #99, opened 2020-01-25
Feature/fetch

Matthijsy:
We used ember-ajax and fetch both in the project. This PR removes all the ajax usage and now uses fetch. For those cases I also changed it from response.then().catch() to async syntax. This makes it more readable.


Matthijsy:
That would be fine to do, but then I would like to use the 'new' mirage style which is currently in PR https://github.com/csvalpha/amber-ui/pull/103

app/controllers/users/forgot-password.js
```diff
@@ -2,33 +2,20 @@ import { inject as service } from '@ember/service';
...
-            break;
-        }
+    async forgotPassword() {
+      const response = await this.fetch.post('/users/reset_password', { body: { email: this.email } });
 
-        this.set('errorMessage', errorMessage);
-      });
+      if (response.ok) {
+        this.set('successMessage', 'Je ontvangt een e-mail met verdere instructies');
+      } else {
```
guidojw: Should ember-fetch/errors' isNotFoundResponse and isServerErrorResponse be used here to preserve the old error messages?

Matthijsy: If we check the code at the backed it will never return a 404 (which it also shouldn't since this gives information about the existing users). So that line of code was never being touched. 

https://github.com/csvalpha/amber-api/blob/staging/app/controllers/v1/users_controller.rb#L50

guidojw: Oh okay, that makes sense!

PR #102, opened 2020-01-26
Update Ember-can

Matthijsy:
### Summary
Time for a little bigger update, Ember-can. The CanMixin is removed from the source project so we need to use the service. For the routes I've injected this globally (since we use it in almost every route) and for the other entities I've injected them manually.

Also the abilities need to be in a singular form now, so I've renamed all the files and merged the cases when they existed both. I did not rename all the occurances, since this is done automatically by ember-can but if we like to do this there as well we could do that.


app/initializers/service-inject.js
```diff
@@ -0,0 +1,7 @@
+export default {
+  name: 'service-inject',
+
+  initialize(application) {
+    application.inject('route', 'can', 'service:can');
```
cmitz: Do we also need this pattern for `session`?

Matthijsy: In the routes we rarely use the session, but we can use it where we want. What do you suggest?

cmitz: Ah, there is my answer. If we barely use it we don't need it here.

We'll see if the need arises in the future when mixins are gone

Matthijsy: We don't need this file actually, but then we need to say `can: service()` if every file we use it (and I don't like that for the routes). So this is actually an auto inject for all routes.

cmitz: I know :)

PR #118, opened 2020-02-28
fixes floating point inaccuracies in debit collections

### LINKED ISSUE #117: debit: Floating point rounding inaccuracies
### How to reproduce
Not entirely sure. It's an issue that happens inconsistently. 
Steps taken were:
1. With the needed permissions, create a new debit collection, upload an appropriately formatted csv that contains several expenses that are not all whole numbers. Note that it is important that you need to have all numbers in the csv rounded to 2 decimals.
2. View the entire collection by going to debit/collections and clicking the collection you just created.

### Expected behaviour
Should, per user in the uploaded csv, show a list of expenses and their descriptions. Should also show a total of the expenses per member. All expenses should be displayed as numbers with at most 2 decimals, as well as the total per user.

### Current behaviour
The expenses are displayed as numbers with at most 2 decimals, but the total expenses per user are sometimes displayed as numbers with an unnecessary amount of decimals:

image

this is an issue, since this becomes misleading and confusing when the total is 0. This can result in scientific notation being used to format a number that, due to floating point inaccuracies resulting from addition, is nearly 0:
image

DrumsnChocolate:
fixes  #117


DrumsnChocolate:
closed by accident. Still, I am curious why buildkite fails

app/controllers/debit/collections/show.js
```diff
@@ -21,6 +21,9 @@ export default Controller.extend({
       transactionsGroupedByUser.transactions.pushObject(transaction);
       transactionsGroupedByUser.totalTransactionAmount += transaction.get('amount');
     });
+    transactions.forEach(transaction => {
```
Matthijsy: I don't think this is the way to go. Firstly we round all the transactions independent, which could cause a difference in the amount that the total says and the real amount we will debit. I think we should do the rounding at the end, so on the total itself. 

Also I do not understand why we multiply by 100 and then devide by 100, could you explain?

Matthijsy:
Buildkite fails on the linter. Don't you have access to buildkite?

app/controllers/debit/collections/show.js
DrumsnChocolate: As for your first comment, We don't round all transactions independently, we round only the totals of every user. 

Every total is multiplied by 100 because this prevents the last 2 decimals from being rounded. Everything that's still a decimal after multiplication by 100 will be rounded, so after division by 100 at most 2 decimals remain. (multiplying by 100 moves the comma by two digits)

DrumsnChocolate:
@Matthijsy it appears I don't. Could you invite/add me?

app/controllers/debit/collections/show.js
DrumsnChocolate: I just realized that I wrote confusing code.

There's two different arrays here, one is an array of per-user-grouped transactions ('transactions', which is initialized as an empty array), the other is 'this.get('model.transactions'), which is a list of all separate expenses.

In hindsight, I see that I made a confusing decision in the variable naming. in line 24, the variable 'transaction' should be named 'transactionsGroupedByUser', since it's not a separate transaction, but one of the elements pushed to the 'transactions' array.

Matthijsy: Thanks for the explanation. How about using JavaScript `toFixed`? This actually does this behavior without the multiplication/dividing mechanism.

 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

DrumsnChocolate: thanks for the suggestion, this is exactly what I needed.

PR #122, opened 2020-03-01
Replace analytics

Matthijsy:
### Summary
Currently we use a really outdated package for google analytics. Since it only requires a few lines to push the current page to google analytics I've decided to do this without a package.


Matthijsy:
Nice catch about the error. Did fix this issue

PR #139, opened 2020-03-25
Remove jquery

Matthijsy:
### Summary
Let's get rid of Jquery! In Ember 4.0 the support will be dropped. I've rewritten the two main components that used it (markdown-editor and ical page). Those are also already transformed to the new Ember class style. 

This has some implications. First of all does the markdown-editor not longer now which text is selected, this means for now that we just apply the new style behind the text. Maybe we can fix this in the future, but is fine for now.

Also the scrollTo function was a Jquery thing. This means that the page-actions and the forum create post do not longer scroll to top/bottom. For now this is probably fine, and we can find a workaround when we improve those components themselves.


wilco375:
I cannot get this to work, it seems `ember-cli-file-picker` also uses jQuery.
image

Also, getting the selection start and end shouldn't be complicated at all without jQuery. If you look at the [HTMLInputElement documentation](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) you can find that it has a `selectionStart` and `selectionEnd` property you can use.

app/controllers/forum/categories/category/threads/thread/posts/index.js
```diff
@@ -42,7 +41,6 @@ export default Controller.extend({
     },
     goToLastPageAndScrollDown() {
       this.set('page', this.totalPages);
-      $('#newForumPost')[0].scrollIntoView(true);
```
wilco375, line 45: Also, scrollIntoView is a native javascript function. So you can just use `document.getElementById('newForumPost').scrollIntoView(true)` here.

app/adapters/application.js
```diff
@@ -27,16 +26,6 @@ export default JSONAPIAdapter.extend(DataAdapterMixin, {
     return pluralize(underscored);
   },
 
-  urlForFindRecord(id, modelName, snapshot) {
-    let url = this._super(...arguments);
-    const query = get(snapshot, 'adapterOptions.query');
-    if (query) {
-      url += `?${$.param(query)}`;
```
wilco375, line 34: I don't know what this function is for, but instead of `$.params` you can use [this](https://dev.to/mornir/jqueryparam-in-vanilla-javascript-59mn).

Matthijsy: I never saw it triggered, do not know for which reason it was build but don't see a reason to keep it

app/controllers/forum/categories/category/threads/thread/posts/index.js
Matthijsy: True, but the scroll down is actually more irritating that useful in my opinion

Matthijsy:
Little update: I did re-enable jquery since we have to replace ember-cli-file-picker. And since that is a bit more work I would like to do that in another PR

Matthijsy:
@wilco375 Forum should be fixed now

PR #141, opened 2020-03-27
Dependency updates

Matthijsy:
### Summary
Another set of dependency updates. The major thing in this PR is the removal of the menu tooltips. Those use jQuery and the newest version is quite broken (or at least does not work for our use case). Since it has little impact i decided to remove it.


wilco375:
I kinda like the popups, it's hard to see with the menu collapsed what's what without them. Isn't there some other way of implementing them?

Matthijsy:
There are probably alternatives, but the current one does not work well for our purpose. Staying at the old version blocks us from removing all jQuery. Would it be an idea to remove them for now and create an issue to re-add them with another library or custom?

PR #153, opened 2020-05-19
Add sponsorkliks alert on main page

Matthijsy:
### Summary
In one of the previous updates of the sponsorkliks extension we made it detectable for Amber. This PR uses this and adds an alert when someone does not have the extension installed. Remarks are welcome :)

Screenshot_2020-05-19_15-12-32


wilco375:
The alignment is off, the left and right sides are not aligned with the articles above

Matthijsy:
Screenshot_2020-05-19_16-37-30

PR #157, opened 2020-05-20
Feature/sponsorkliks alert update

Matthijsy:
Make the sponsorkliks alert closeable


wilco375:
You are not checking if the user is on a mobile browser yet, are you?

Matthijsy:
No, that was quite hard (to do good) and since you can dismiss the alert is
it such a problem?



Op wo 20 mei 2020 om 14:08 schreef Wilco <notifications@github.com>:

wilco375:
Not really, just would be nice

PR #159, opened 2020-05-25
Fix dropzone

### LINKED ISSUE #150: Image dropzone returns 401 when uploading images
### How to reproduce
Login with a account with image creation rights
1. Upload a new image to a album

### Expected behaviour
The image is added to the album

### Current behaviour
The upload dropzone returns a 401 and a error is shown to the user
Screenshot 2020-04-23 at 13 48 17

wilco375:
### Summary
Fixes #150 
Also, there was no preview shown of the uploaded images because it is shown as a base64 encoded data string which was not allowed by cors policy. That is also fixed with this PR.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -23,7 +23,7 @@ module.exports = function(environment) {
       'script-src': '\'self\' www.google-analytics.com www.googletagmanager.com',
       'font-src': '\'self\' fonts.gstatic.com',
       'connect-src': '\'self\' sentry.io',
-      'img-src': '\'self\' camo.csvalpha.nl www.google-analytics.com img.youtube.com data:',
+      'img-src': '\'self\' camo.csvalpha.nl www.google-analytics.com img.youtube.com',
       'style-src': '\'self\' \'unsafe-inline\' fonts.googleapis.com/',
       'media-src': '\'self\'',
       'object-src': '\'self\'',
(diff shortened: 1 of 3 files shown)


config/environment.js
```diff
@@ -23,7 +23,7 @@ module.exports = function(environment) {
       'script-src': '\'self\' www.google-analytics.com www.googletagmanager.com',
       'font-src': '\'self\' fonts.gstatic.com',
       'connect-src': '\'self\' sentry.io',
-      'img-src': '\'self\' camo.csvalpha.nl www.google-analytics.com img.youtube.com',
+      'img-src': '\'self\' camo.csvalpha.nl www.google-analytics.com img.youtube.com data:',
```
Matthijsy, line 26: I am not sure if I like this, this is potentially insecure.

```
data: Allows data: URIs to be used as a content source. This is insecure; an attacker can also inject arbitrary data: URIs. Use this sparingly and definitely not for scripts.
```
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src

wilco375: Hmm that's true... Is there any other way to preview the uploaded images, without adding :data to the cors policy?

Matthijsy: Maybe blobs? Or check with dropzone itself what the recommend

wilco375: Well, it is a blob, which is set as the image source as a base64 encoded string. Don't know if there's any other way to set the image source from a blob, but then you basically have the same problem again as described on that site, don't you?

Matthijsy: I have created a issue with the package, wondering how other developers handle with this. https://github.com/FutoRicky/ember-cli-dropzonejs/issues/112

wilco375: Great, we'll see what they'll respond with. For the time being, I removed :data from the CSP. For now we'll then at least have a working dropzone, just without preview images.

PR #160, opened 2020-05-25
Add about alpha videos

wilco375:
### Summary
This PR adds 'Over C.S.V. Alpha' to the external front page.

image
image

Also see https://github.com/csvalpha/amber-api/pull/161


app/abilities/about-alpha-video.js
```diff
@@ -0,0 +1,7 @@
+import { inject as service } from '@ember/service';
+import { Ability } from 'ember-can';
+
+export default Ability.extend({
```
Matthijsy, line 4: There is no about-alpha-videos page right? Then we don't need a ability I think?

wilco375: Okay then I'll remove that. Don't know enough about Ember yet so I just copy pasted some stuff 😅

app/templates/components/about-alpha.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="about-alpha__container">
+  {{#each videos as |video|}}
+    <a class="about-alpha__video" href="https://youtu.be/{{video.videoId}}" target="_blank" title="{{video.title}}">
```
Matthijsy, line 3: Could you use the YoutubeVideo component? That does most of this things for you and does some smart things with the thumbnail

wilco375: You don't want to play the video in an iframe when it is that small

app/components/youtube-videos.js
```diff
@@ -0,0 +1,36 @@
...
+    title: 'Over C.S.V. Alpha #3 Christine vertelt over commissies'
+  }
+]
+
+export default Component.extend({
+  store: service(),
+  mainVideo: videos[0],
+  sideVideos: computed('videos', 'mainVideo', function() {
+    let mainVideo = this.mainVideo
+    return videos.filter(video => video.videoId !== mainVideo.videoId)
```
Matthijsy, line 29: My bad but I think it can be even cleaner:
```
 return videos.filter(video => video.videoId !== this.mainVideo.videoId)
```
And the remove the previous line

app/templates/components/youtube-videos.hbs
```diff
@@ -0,0 +1,16 @@
+<div class="row">
+  <div class="col-xs col-lg-9">
+    <YoutubeVideo @youtubeId="{{mainVideo.videoId}}" @title="{{mainVideo.title}}" />
+  </div>
+  <div class="col-xs col-lg-3">
+    <div class="about-alpha">
+      <div class="about-alpha__container">
+        {{#each sideVideos as |video|}}
+          <a class="about-alpha__video" href="https://youtu.be/{{video.videoId}}" target="_blank" title="{{video.title}}" {{action "changeVideo" video}}>
```
Matthijsy, line 9: I don't think it needs to be a link anymore since we do not link to youtube anymore

app/components/youtube-videos.js
```diff
@@ -0,0 +1,36 @@
+import { inject as service } from '@ember/service';
```
Matthijsy, line 1: Maybe place this file within the 'index' directory in the components. Since it is related to the index page only

PR #162, opened 2020-05-29
Update of dependencies

Matthijsy:
### Summary
Time for some more updates, this time mainly eslint updates. For now `ember/no-get` is disabled since we use it everywhere, but we want to enable this at some point in time.


Matthijsy:
No more outdated packages :))

Screenshot_2020-05-29_15-59-49

PR #167, opened 2020-06-10
Remove this.get('property') in favor of this.property

Matthijsy:
### Summary
For most calls we don't need `this.get('')` anymore, we simply do `this.property`. This PR removes all the cases. The only issue is that `this.property.something` could fail when property is undefined. However, that is actually quite handy since I already found some bugs that did not raise an error due to the `.get()`. Hopefully there are no bugs but could be that this introduces some.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,9 +38,7 @@ module.exports = {
     'ember-suave/require-access-in-comments': 0,
     'ember/no-jquery': 1,
     'ember/no-mixins': 1,
-    'lines-between-class-members': ['error', 'always', { 'exceptAfterSingleLine': true }],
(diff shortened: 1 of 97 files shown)


wilco375:
image
The photo album page is still not behaving as it should

wilco375:
Also there seems to be some loop on a forum thread page (/forum/categories/[category]/threads/[thread]/posts?page=1) that causes the page not to respond.

wilco375:
The single poll page is also broken
image

wilco375:
Also the 'SOG namen leren' page breaks when you click the start button 
image

wilco375 (approved):

PR #169, opened 2020-06-10
Feature/octane

Matthijsy:
### Summary
Time for even more octange changes! This time it is time for the models. 

I also tried to cleanup the models a bit. The only models that could use some more cleanup is the form models, but those are quite complicated and probably need some refactor all together, so will do that later on.


app/models/activity.js
```diff
@@ -1,39 +1,34 @@
...
+  @attr coverPhoto;
+  @attr coverPhotoUrl;
+
+  // Relations
+  @belongsTo('user') author;
+  @belongsTo('group') group;
+  @belongsTo('form/form') form;
 
-  // Computed properties
-  formattedPrice: computed('price', function() {
```
wilco375, line 27: Why is the price not formatted to two decimals anymore?

app/models/forum/post.js
```diff
@@ -1,19 +1,21 @@
...
+import Model, { belongsTo, attr } from '@ember-data/model';
 
-export default Model.extend({
+export default class Post extends Model {
   // Properties
-  message: attr('string'),
-  messageCamofied: attr('string'),
-  createdAt: attr('date'),
-  updatedAt: attr('date'),
+  // @attr message;
```
wilco375, line 6: Why is this attribute commented out?

app/models/group.js
```diff
@@ -33,16 +31,21 @@ export default Model.extend(AvatarModelMixin, {
...
 
     return null;
-  }),
-  alternativeGroupLogoText: computed('name', function() {
-    return `Het logo van ${this.name}`;
-  }),
+  }
+
+  get avatarUrlOrDefault() {
+    return this.avatarUrl || '/images/fallback/group_avatar_default.png';
```
wilco375, line 37: Why did you not make this a constant like the other default images?

app/models/forum/post.js
```diff
@@ -1,19 +1,21 @@
...
   // Relations
-  author: belongsTo('user'),
-  thread: belongsTo('forum/thread'),
+  @belongsTo('user') author;
+
+  @belongsTo('forum/thread') thread;
 
   // Computed properties
-  isUpdated: computed('createdAt', 'updatedAt', function() {
+  @computed('createdAt', 'updatedAt')
```
wilco375, line 17: (Why) do you need to explicitly give these variables here? Isn't it the same as, for example, `userIsCurrentlyMember` in membership.js?

app/models/group.js
Matthijsy: because this image is only used for groups, the other are (as far as I remember) used at multiple locations

wilco375 (approved):

PR #171, opened 2020-06-10
Drop mail enrolled route

Matthijsy:
### Summary
The mail enrolled function on activities is already quite buggy since the start. I would like to remove it, and later on implement it in a better way (using a temporary mail alias).


wilco375:
I do not agree with completely dropping this for now. It is a lot of work to manually copy and paste all users you want to send an email to. I would at least add a popup or something from which you can copy a list of all the email addresses of enrolled users. That way you can easily send an email from your own mail client. That might even be good enough to not have to reimplement this at all.

Matthijsy:
We decided not to do that since a lot of users will probably drop all those users into the 'to' field which creates a data breach. If this tool is still usefull we can leave it here until we re-implemented it

wilco375:
Fair point. In that case I would just leave it as is for now. I still use it sometimes and I see other people using it as well.

PR #172, opened 2020-06-12
Feature/Article pinning

guidojw:
This PR implements support for article pinning. The article sections on the homepage and the article page will now show a pinned article (if one exists) first. An article can be pinned by going to the edit article page.

This PR needs https://github.com/csvalpha/amber-api/pull/165


app/routes/articles/index.js
```diff
@@ -24,5 +24,11 @@ export default IndexRouteUnauthenticated.extend(PagedModelRouteMixin, {
         canAccess: this.can.can('create articles')
       }
     ];
-  })
+  }),
+
+  model(params) {
+    params.paramMapping = this.paramMapping;
+    params.sort = '-pinned,-created_at';
```
Matthijsy, line 31: This breaks the sorting bar probably. So I think we should do something like this `params.sort = '-pinned' + params.sort`

PR #174, opened 2020-06-22
Add first page and last page buttons to pagination.

### LINKED ISSUE #165: No paginations first and last buttons
### How to reproduce
Go to a long forum thread

### Expected behaviour
There is a button to skip to the last page

### Current behaviour
There is only a button to go to the next page

RubenSmit:
- Add first page and last page buttons to pagination.
- Show only 5 pages on mobile
Fixes #165

### DIFF
--- a/app/components/page-numbers.js
+++ b/app/components/page-numbers.js
@@ -1,0 +1,6 @@
+import PageNumbersComponent from 'ember-cli-pagination/components/page-numbers';
+
+export default PageNumbersComponent.extend({
+  numPagesToShow: 5,
+  showFL: true
+});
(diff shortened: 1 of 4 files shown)


Matthijsy:
The page numbers package that we use actually already has support for this, but we removed the configuration for this by accident. Restoring this file should fix it https://github.com/csvalpha/amber-ui/commit/b9c125fc2b3c6f08c913fcbf83de53d1042d4b7a#diff-06fb4a4673a869b1ee7e9260347e7f26L6

RubenSmit:
Whahaha, what a waste of time then...

wilco375:
Do we also want to add this code:
```
  actions: {
    pageClicked(number) {
      this._super(number);
      window.scrollTo(0,0);
    }
  }```

This will scroll the page to the top after changing to a different page. I personally quite like this feature but I don't know how you guys think about it?

RubenSmit:
We do not need that since it is already default behaviour when clicking links.

PR #204, opened 2020-09-10
Fix notification disappearing when one hour or minute is 0 instead of all

wilco375:


Matthijsy (commented):
:open_mouth: good one :+1:

Matthijsy (approved):

PR #205, opened 2020-09-11
Improve front page buttons

wilco375:
### Summary
- Improve responsiveness of buttons on public site
- Add button that links to the sign up form

### Other information
This year, we are also allowing people to become a member by filling out and emailing a PDF because of all the online activities. I added a button that makes it a bit easier to find the info page that has this PDF on it.

image


Matthijsy (approved):
Looks good! How about making a Google forms or something similar instead of a PDF? Would make it even better

wilco375:
> Looks good! How about making a Google forms or something similar instead of a PDF? Would make it even better

The problem is that (I think) we need a signature. For now this works. Might be something to look into for next year maybe though.

PR #209, opened 2020-09-14
Fix model relations and date attributes

wilco375:
### Summary
There were several issues with the last deploy. This PR fixes that: 
- Users could not be added to a group because of an incomplete model relation
- Date attributes were not parsed as dates. This caused user group filtering to be broken and may have caused other issues elsewhere

### DIFF
--- a/app/models/activity.js
+++ b/app/models/activity.js
@@ -12,8 +12,8 @@ export default class Activity extends Model {
   @attr startTime;
   @attr endTime;
   @attr('capitalized-string') category;
-  @attr createdAt;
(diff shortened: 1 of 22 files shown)


Matthijsy (approved):
Code looks good! did you do a ctrl+f on all `at` and `date` attributes in the model directory? Then we probably did not miss any

wilco375:
> Code looks good! did you do a ctrl+f on all `at` and `date` attributes in the model directory? Then we probably did not miss any

Yes, I did a regex search and replace on that 😉

PR #222, opened 2020-10-18
Make FaIcon use anglebrackets

Matthijsy:
### Summary
This PR removes a dep warning about fa-icon, by using the new anglebrackets


wilco375 (changes requested):
The `prefix` arguments are missing an `@`

wilco375 (approved):

PR #230, opened 2020-11-20
Fix edit button of article, activity and poll

### LINKED ISSUE #173: Edit article/activity button is always visible
### How to reproduce
Visit the page of any article/activity.

### Expected behaviour 
Only articles for which I have editing permissions should show the "Edit article" button.

### Current behaviour
Every article/activity shows the button.
Trying to edit these resources returns a HTTP 403, so that works.

wilco375:
Fixes #173


DrumsnChocolate (approved):
completely nice.
Checked for consistency, and recommended some changes which were subsequently committed by me or @wilco375

PR #235, opened 2020-11-21
Feature/Allow to run dev server via Docker

guidojw:
This PR adds a Dockerfile and a Docker Compose configuration that allows you to run the project without the need to install all dependencies; Docker will do this for you.

The docker-compose.development.yml file connects to the external Docker network named "amber_development_default". This network will be automatically created by the API if you change COMPOSE_PROJECT_NAME in .env on amber-api from amber_\<env\> to amber_development.


README.md
```diff
@@ -4,35 +4,45 @@ Alpha AMBER UI
...
+*: Installing the Ember CLI is required if you're not using Docker, but it's also useful with Docker if you want to make use of its features like Code Generators.
 
-* `git clone <repository-url>` this repository
-* change into the new directory
-* `yarn add` (or `npm install`)
+## Installation
+### With Docker
+* Build the project using `docker-compose -f docker-compose.development.yml build ui`. This will install the dependencies and set up the image. If dependencies are updated/added, you need to run this command again.
+
+Tip: to run commands in a container, you need to find the container's ID using the `docker ps` command first. Copy the ID of the container with image "amber_development_api".
```
Matthijsy, line 22: Do we need this here? Only relevant for API I think?

guidojw: The part about running commands in containers? I think that's useful information here as well for example for running the linter locally

guidojw:
I can't get live reloading to work with the Docker volumes but the rest seems to work fine

PR #236, opened 2020-11-21
Remove jQuery

wilco375:
### Summary
This PR removes jQuery and all uses of it.

See #22


Matthijsy (approved):
Nice! Always good to have less in vendor

PR #237, opened 2020-11-21
Implement new file picker without jQuery

wilco375:
PR #236 caused the file picker to break since that depended on jQuery. This PR implements ember-file-upload which does not require jQuery.


Matthijsy (approved):
Nice work!

PR #239, opened 2020-11-21
Feature/Dark theme

### LINKED ISSUE #175: Dark theme
In order to save our precious eyes I thought it would be nice if we had our very own dark mode. I've been looking into Ember addons for dark mode and found there's already a very useful addon for this (ember-theme-changer). 

I was wondering if our branding allows this? If it does I would be more than happy to further look into this.

guidojw:
This PR implements a dark theme for the website.

I have taken out all the relevant color css properties of the original style and put them in their own dark|light folder in app/styles/themes, in the same structure as original.

This PR resolves #175


guidojw:
Might revisit later

PR #240, opened 2020-11-21
Upgrades

Matthijsy:
### Summary
A few yarn upgrades.

For the review: please verify that the autoreload from Ember is working. I have some issues with it but probably my local machine


wilco375 (approved):
Looks good, autoreload also seems to work fine for me

PR #241, opened 2020-11-21
Remove users without thumbnail from list, improve difficulty 1 layout

wilco375:
### Summary
This PR improves the layout of the Easy name trainer level by showing the images in a 2x2 grid instead of a list.
Additionally, users without an avatar image are now filtered from the trainer.


Matthijsy (approved):
Nice!

PR #251, opened 2020-12-10
Improve sentry logging

Matthijsy:
### Summary
This will make use of the current way to access sentry, have production and staging within one project (with 2 environments) and use the 'releases' feature of sentry indicating when the issue was introduced.


config/environment.js
```diff
@@ -201,22 +194,24 @@ module.exports = function(environment) {
...
-
     ENV.APP.rootElement = '#ember-testing';
     ENV.APP.flashNoticeDefaultDuration = 1;
   }
 
+  ENV['@sentry/ember'] = {
+    sentry: {
+      dsn: 'https://invalid@xx.ingest.sentry.io/12345', // invalid key, will be replaced when run as prod
+      environment: deployTarget,
+      release: `amber-ui@${  process.env.BUILD_HASH}`
```
guidojw, line 205:
```suggestion
      release: `amber-ui@${process.env.BUILD_HASH}`
```

Dockerfile
```diff
@@ -2,13 +2,14 @@ FROM madnificent/ember:3.17.0 as ember
 MAINTAINER C.S.V. Alpha <ict@csvalpha.nl>
 
 ARG DEPLOY_TARGET='production'
+ARG BUILD_HASH='unkown'
```
guidojw, line 5:
```suggestion
ARG BUILD_HASH='unknown'
```

guidojw (changes requested):
Rest looks good 👍🏼

Dockerfile
```diff
@@ -2,13 +2,14 @@ FROM madnificent/ember:3.17.0 as ember
 MAINTAINER C.S.V. Alpha <ict@csvalpha.nl>
 
 ARG DEPLOY_TARGET='production'
+ARG BUILD_HASH='unknown'
```
guidojw, line 5: Shouldn't these (also DEPLOY_TARGET) be ENVs because they're environment variables in the docker-compose.buildkite.yml too?

Dockerfile
```diff
@@ -2,13 +2,14 @@ FROM madnificent/ember:3.17.0 as ember
...
 
 COPY package.json /app/package.json
 COPY yarn.lock /app/yarn.lock
 RUN yarn install --ignore-engines
 
 COPY . /app
-RUN DEPLOY_TARGET=$DEPLOY_TARGET ember build --environment=production
+RUN DEPLOY_TARGET=$DEPLOY_TARGET BUILD_HASH=$BUILD_HASH ember build --environment=production
```
guidojw, line 12: That way the arguments here can be omitted I think

docker-compose.buildkite.yml
```diff
@@ -8,6 +8,11 @@ services:
 
   app-production: &production
     build: .
+    environment:
+      - BUILD_HASH=${BUILDKITE_COMMIT}
```
guidojw, line 12: The question in Dockerfile **or** should these be made build args (and in app-staging)?

Dockerfile
Matthijsy: What we do is set it as environment in docker-compose for buildkite, and then use that environment variable in the build step for this docker image. The difference is that a build argument is only used during build, where the environment variable is used and set during execution

guidojw: That makes a ton of sense now, thanks for the explanation 👍🏼

guidojw (approved):

PR #252, opened 2020-12-10
Only show status dot when with permission

### LINKED ISSUE #249: Forms are loaded when without permission
### How to reproduce
1. Login as 'oud-lid'
1. Go to the main page

### Expected behaviour
No requests are done for things related to 'forms' (since you do not have this permission)

### Current behaviour
Some forms are loaded, resulting in 403 errors

This does not break anything but does give some errors so would be nice to fix
Related sentry issue: https://sentry.io/organizations/csvalpha/issues/1590968950/?project=186017&query=is%3Aunresolved&sort=freq&statsPeriod=14d

Matthijsy:
### Summary
The form of an activity was loaded on the frontpage for the status dot, but also when without permission. This PR fixes that.

Fixes #249


app/templates/components/tools/upcoming-activities.hbs
```diff
@@ -20,11 +20,12 @@
   </div>
 
   <ul class="list-group list-group-flush">
+
```
guidojw, line 23: Did you mean to add this new line?

guidojw (approved):
looks good to me

app/templates/components/tools/upcoming-activities.hbs
Matthijsy: Not on purpose I think

PR #260, opened 2021-02-18
Feature/dep update

Matthijsy:
### Summary
Time for even more updates! We have a lot of old codestyles that still need to be upgraded, for that I have set the linter to warn instead of error, since we should update it first.


wilco375 (approved):
Looks good

PR #262, opened 2021-02-20
Update ember auth

Matthijsy:
### Summary
Ember simple auth doesn't need mixins anymore, so this PR updates all routes to the new ocatane style with the use of the simple auth mixins!


wilco375 (approved):
Did not read all 151 changed files but everything seems to be working as expected.

PR #264, opened 2021-02-21
New pagination style

Matthijsy:
### Summary
Currently, we use ember-cli-pagination, however this relies on a mixin to work. We would like to have an application without mixins, since this is the new ember-way of doing things. So i removed the plugin, and extended the ember data store with an `queryPaged` method, that adds some meta to work in the `PageNumbers` component that then can show the pages that are available and navigate between those.


guidojw (approved):
Nice changes, works well!

PR #266, opened 2021-02-23
Advertisement links

wilco375:
### Summary
Add new advertisement links

image

### DIFF
--- a/app/styles/components/advertisement-tool.scss
+++ b/app/styles/components/advertisement-tool.scss
@@ -1,5 +1,5 @@
 // CSS3 code for animation of advertisements
-$amount-of-advertisements: 9;
+$amount-of-advertisements: 10;
 $single-advertisement-width: 175px;
 $advertisement-animation-duration: 15s;
 $advertisement-animation-holder-width: - $amount-of-advertisements * $single-advertisement-width;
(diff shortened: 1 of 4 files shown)


app/styles/components/advertisement-tool.scss
```diff
@@ -1,5 +1,5 @@
 // CSS3 code for animation of advertisements
-$amount-of-advertisements: 9;
+$amount-of-advertisements: 10;
```
Matthijsy, line 2: You added 3 new advertisements right? So then it should be 12

Matthijsy: Ah i now see that it is 1 advertisement containing of multiple sub-adverstiements

app/templates/components/advertisement-tool.hbs
```diff
@@ -2,9 +2,17 @@
   <div class="tool adv-container">
     <div class="d-flex adv-row align-items-center">
       {{#each advertisements as |advertisement index|}}
-        <a class="advertisement-item col-auto" href={{advertisement.link}} title={{advertisement.title}} target="_blank">
-          <img src={{advertisement.image}} alt={{if advertisement.alt advertisement.alt advertisement.title}}>
-        </a>
+        {{#if advertisement.links}}
+          <div class="advertisement-item advertisement-item--links col-auto" href={{advertisement.link}} title={{advertisement.title}} target="_blank">
```
Matthijsy, line 6: I think the href and titlte are not needed here, since those are not defined in the object

wilco375: Oops, you are right. Resolved.

Matthijsy (approved):

PR #268, opened 2021-02-24
Fix thread permission bug

### LINKED ISSUE #267: Cannot reach forum without 'admin' permissions
Currently the members cannot reacht the forum, probably one of the `canAccess` methods in one of the forum routes are returning false for them. The ICT committee can still reach it, so probably it is something with a permission that we do have and members do not

guidojw:
#262 introduced a regression where a wrong permission was being used for showing threads.

This PR resolves #267


Matthijsy (approved):
Nice catch!

RubenSmit:
Nice! Just wondering if it has already been deployed. The problem still persists for me.

PR #274, opened 2021-02-27
Fix/Make OauthAuthorizeRoute an AuthenticatedRoute

### LINKED ISSUE #271: OAuth route not authorized
### How to reproduce
1. Go to streep.csvalpha.nl while you are logged out of csvalpha.nl

### Expected behaviour
Streep.csvalpha.nl redirects you to csvalpha.nl, which redirets you to the login page, since you first need to authenticate

### Current behaviour
You get the 'Oauth authorize' screen, but it does not work since you are not authorized at the moment

guidojw:
This PR makes the route for authorizing OAuth applications extend the AuthenticatedRoute.

This PR resolves #271


Matthijsy (approved):
Nice fix!

Issue #272, opened 2021-02-26
Navigating through polls doesn't work


wilco375:
### How to reproduce
1. Go to a poll
2. Click on polls in the breadcrumb or menu to go back to the overview
3. Go to another poll

### Expected behaviour
The poll is shown

### Current behaviour
An error is thrown
![image](https://user-images.githubusercontent.com/7385023/109293698-7282c900-782c-11eb-8eb5-3d6da82848a8.png)

Matthijsy:
Probably related to https://github.com/csvalpha/amber-ui/issues/250

PR #275, opened 2021-02-27
Fix/Poll ability checking for memberships

### LINKED ISSUE #250: Poll 'isOwner' gives error
### How to reproduce
1. Login (as member)
2. Load a poll

Now an error occurs (at least sentry said so). I cannot reproduce it with my account so probably only for members. I suspect that the issue is that a poll cannot be attached to a group, so the 'isOwner' method probably does not make any sense

Sentry link: https://sentry.io/organizations/csvalpha/issues/1893358502/?project=186017&query=is%3Aunresolved&sort=freq&statsPeriod=14d

guidojw:
This fixes the bug where the poll.isOwner method checked if a user is member of the group the poll is made by, while a poll can be made by users only.

This PR resolves #250 & #272 
fixes AMBER-UI-PRODUCTION-TG

### DIFF
--- a/app/abilities/poll.js
+++ b/app/abilities/poll.js
@@ -14,11 +14,11 @@ export default Ability.extend({
   canDestroy: computed('session.currentUser', function() {
     return this.session.hasPermission('poll.destroy');
   }),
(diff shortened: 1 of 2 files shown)


Matthijsy (approved):
Nice!

PR #276, opened 2021-02-27
Fix/Readd layoutManager.closeProfileMenu

### LINKED ISSUE #269: Going to a new page via profileMenu does refresh application
### How to reproduce
1. Go the the website
2. Open the profile menu
3. Click on a link

### Expected behaviour
The page is loaded, and the profile menu is closed.

### Current behaviour
The full application refreshes. 

It seems to be related to a call to `layoutMananger.closeProfileMenu`, this method does not exists anymore. probably we have to change this call to `layoutManager.toggleProfileMenu`. 

Sentry: https://sentry.io/organizations/csvalpha/issues/2233600524/?project=186017&query=is%3Aunresolved

guidojw:
#260 removed the layoutManager.closeProfileMenu method. This PR makes the menu auto closing use the toggle method instead.

This PR resolves #269 
fixes AMBER-UI-PRODUCTION-XF


app/components/menu-sidebar-item.js
```diff
@@ -6,8 +6,8 @@ const MenuSidebarItemComponent = Component.extend({
   minor: false,
   actions: {
     closeAfterNavigation() {
-      if (this.minor) {
-        this.layoutManager.closeProfileMenu();
+      if (this.minor && this.layoutManager.profileMenuOpen) {
```
Matthijsy, line 9: I know I suggested this myself, but maybe it is better to add the `closeProfileMenu` method again, since now we have to check the status of the profileMenu here before toggling, otherwise we do not have to care about the state of the service and just say close.

guidojw: Okay, I'll add it back 👍🏼

Matthijsy (approved):

PR #282, opened 2021-03-01
Make the abilities use native classes

Matthijsy:
### Summary
This PR migrates all the abilities to use native classes instead of classic ember classes


guidojw (approved):
👍🏼

PR #287, opened 2021-03-02
Fix abort error on oauth after login

### LINKED ISSUE #281: OAuth authorize route auto invalidates session when logging in before authorizing
With #274 you now get redirected to the login page if you're not authenticated but the rest still goes wrong.

### How to reproduce
1. Go to streep.csvalpha.nl when not logged in on csvalpha.nl
2. Click "Sign in with C.S.V. Alpha"
3. Login on the main site

### Expected behaviour
It should redirect you to streep.csvalpha.nl where you are now logged in.

### Current behaviour
You get redirected to the public page of csvalpha.nl where you're still logged out.

After looking through the network requests of my browser console, I found this:
Screenshot_1547

A request that's missing here but is executed when authorizing SOFIA from a logged in account is GET /users/:id/permissions, executed by [this line](https://github.com/csvalpha/amber-ui/blob/64e18bc883a4cee73d8243479c60858e4690071c/app/services/session.js#L34).
I'm not sure why this line isn't being executed but it probably causes the loadUser method to throw which then automatically [invalidates the session](https://github.com/csvalpha/amber-ui/blob/64e18bc883a4cee73d8243479c60858e4690071c/app/services/session.js#L22) (last 2 requests in the screenshot).

Matthijsy:
### Summary
Fixes #281 

The Oauth route does an redirect as soon as it sees that the current user has already authorized the application. However, this throws an 'abortError' in ember, which just indicates that the current transition has been aborted. This however combined with the loadUser caused it to invalidate. this PR fixes that by checking that the error is not an abortError

### DIFF
--- a/app/services/session.js
+++ b/app/services/session.js
@@ -19,7 +19,9 @@ export default class SessionService extends BaseSessionService {
     try {
       await this.loadUser();
     } catch(err) {
-      await this.invalidate();
+      if (err.name !== 'AbortError') {
+        await this.invalidate();
+      }
     }
   }
 


guidojw (approved):
🤯 Makes sense now why some of the requests weren't executed, they were aborted 
Looks good!

PR #310, opened 2021-05-17
Hotfix for form signup

wilco375:
### Summary
The optionIds are updated later than updateMultipleChoiceAnswers is called, causing selected form options to lag behind. The form elements should be rewritten but for now, adding a small timeout works as a hotfix.


Matthijsy (approved):
Dirty but works, probably will stay here too long but fine for now

PR #315, opened 2021-05-20
feat: send clientId with OAuth2 password grant

guidojw:
Allows for a way to set the `clientId` sent with OAuth requests, since this is necessary with Doorkeeper v5.5.0.
This requires public (confidential: false) Doorkeeper clients to be made on the API for both `staging` and `production`, which the can be set in `environment.js`.

I added the `.env.example` and `ember-cli-dotenv` to make setting the `clientId` in development environments easier.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,1 @@
+OAUTH_CLIENT_ID=<id>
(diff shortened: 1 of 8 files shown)


Matthijsy:
If you create the applications on staging and prod we can test it out on staging, the code seems fine

Matthijsy (approved):
Let's try this out on staging!

PR #317, opened 2021-05-22
Issue/forum threads sorting

### LINKED ISSUE #125: Forum threads in categories are sorted from old to new
It seems more logical to me that the most recently updated forum thread is on top and the oldest one on the bottom in on the category page, but the forum thread with the latest update is shown last now.

DrumsnChocolate:
fixes #125
a09c2ca3085e16d6e3949291154da80fd7cf2823 is the relevant commit, see commit message for explanation:
>Took a peak at how routes/forum/categories/category/threads/thread/posts/index handles the loading of the model, since every post belongs to one thread, similar to how every thread belongs to one category. Applying the same technique for querying based on a filter in params was pretty straightforward, but required the addition of a filter in the backend. See the equally named branch in amber-api for this addition.
Note: the params attribute in the threads/index route is somewhat different from the params in the posts route; the corresponding posts controller does not extend filterableAndSearchable, whereas the threads controller does extend that mixin.

Requires https://github.com/csvalpha/amber-api/pull/226


app/routes/forum/categories/category/threads/index.js
```diff
@@ -38,7 +39,15 @@ export default class ThreadIndexRoute extends AuthenticatedRoute {
...
-    return this.modelFor('forum.categories.category');
+  async model(params) {
+    const category = this.modelFor('forum.categories.category');
+    assign(params, {
+      filter: { category: category.id }
+    });
+    const threadsPromise = await this.store.queryPaged('forum/thread', params);
+    return {
+      category,
+      threads: threadsPromise
```
Matthijsy, line 50: Where is this data used? I only see a change for the model, but not a change in how the model works.

DrumsnChocolate: I believe this is necessary because queryPaged is passed `params`, which if I recall correctly is modified by (or based on) the FilterableAndSortableMixin. If you look at the controller for this route, https://github.com/csvalpha/amber-ui/blob/staging/app/controllers/forum/categories/category/threads/index.js you can see that the queryParams are set to ['search', 'sort'] these are defined in the FilterableAndSortableMixin, and the way queryParams work is defined here: https://guides.emberjs.com/release/routing/query-params/
In the template that belongs to the this route, the search and sort parameters are adjustable for the user through the FilterBar component.
Without passing this code, the `params` are never used in the query to the backend, and the threads that belong to the category are neither sorted nor filtered.

wilco375 (approved):
Seems to be working

PR #318, opened 2021-05-22
Issue/activity question order

### LINKED ISSUE #304: Can't change order of questions for activity registration
### How to reproduce
1. Create a activity with some questions
2. Try to reorder the questions using "omhoog" or "omlaag"

### Expected behaviour
The order of the questions should change

### Current behaviour
Nothing happens

DrumsnChocolate:
fixes #304
The gist is that methods like `lastObject` should not be called on 'plain' js arrays. When calling `toArray` on an `EmberArray`, it will turn into a 'plain' js array. However, ember augments these types of arrays (I'm not clear on exactly how this works), such that `lastObject` is still callable on the new array. However, when calling `addObjects` on the plain js array, the `lastObject` won't update, but rather point to the same element in the array as before. 
Or at least that's what seemed to be the problem since the issue has been resolved by copying `EmberArrays` using `A()` and then modifying that copy.
Why did we move away from using `@union` for taking the union of two ember arrays?


Matthijsy:
`@union` is a computed property which are deprecated if I am right. The correct way to do that now is using a getter/setter method if I am right

DrumsnChocolate:
https://api.emberjs.com/ember/release/functions/@ember%2Fobject%2Fcomputed/union I don't see it mentioned here

app/components/form/closed-question.js
```diff
@@ -5,6 +5,10 @@ import OpenQuestionComponent from './open-question';
 export default OpenQuestionComponent.extend({
   store: service(),
   questionTypes: ClosedQuestionTypes,
+  init() {
+    this._super();
+    this.send('addOption');
```
Matthijsy, line 10: What does this do? This is only run once, not every time this component is rendered, so I am not sure if this does what it is supposed to do

app/models/form/closed-question.js
```diff
@@ -22,7 +22,7 @@ export default class ClosedQuestion extends Model {
   }
 
   get sortedOptions() {
-    return this.options.toArray().sort((a) => a.position);
+    return this.options.sortBy('position');
```
Matthijsy, line 25: Isn't that a JQuery function?

app/components/form/closed-question.js
DrumsnChocolate: this is done such that the closed question type always starts with one dummy/example option.

app/models/form/closed-question.js
DrumsnChocolate: ehm, I don't know? is it? https://api.emberjs.com/ember/release/classes/EmberArray shows it as a normal method for ember

wilco375 (approved):

PR #338, opened 2021-08-29
Fix filter bar

wilco375:
### Summary
The filter bar was not shown because BsDropdown seemed to have broken. Updating ember-bootstrap seems to have fixed the issue. Additionally, I have fixed the alignment of the search bar.


Matthijsy (approved):
Did not test, but seems good

PR #339, opened 2021-08-29
Fix activity consumption list

### LINKED ISSUE #325: Cannot show/print the "streeplijst"
### How to reproduce
1. Go towards an activity with some enrollments
2. Click on "print ingeschreven"
3. Click on "streeplijst"

### Expected behaviour
The list below should change to an "streeplijst" that can be printed

### Current behaviour
Nothing happens

wilco375:
### Summary
Fixes #325


app/components/form/consumption-list.js
```diff
@@ -0,0 +1,12 @@
+import Component from '@ember/component';
+
+const ConsumptionListComponent = Component.extend({
```
Matthijsy, line 3: Could you write this component in the new octane style? Saves some migrations later

wilco375: Unfortunately this is not doable without updating a ton of other components too

DrumsnChocolate (approved):

PR #344, opened 2021-09-12
Add new advertisements

wilco375:
### Summary
Adds new advertisements to the advertisement tool. Because of the amount of advertisements added, the layout has been adapted to allow two advertisements to show below each other.

image


guidojw (approved):
Can't test right now but looks good

PR #358, opened 2021-10-15
fix: change to compatible node versions

guidojw:
### Summary
Changes the Node.js version in .node-version to a version compatible with the Ember version we're currently using. Did this because on my laptop `yarn  install` didn't work on Node 16 and on my pc `ember s` somehow didn't work. Reverting back to an older version of Node.js fixed this.
Also renamed `.node-version` to `.nvmrc` because my NVM didn't see the old file and I figured everyone is using either no version manager or NVM (is that true?).

See [Ember-Node support](https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md)
(The file used to be called `.nvmrc` a long time ago: https://github.com/csvalpha/alpha-banana-ui/pull/1341)

TODO:

- [x] Disable engine updates on depfu dashboard (or make them non-major only).


DrumsnChocolate (approved):
yarn install works, so I presume npm also still functions.

guidojw:
Engine updates on depfu are now minor only

PR #384, opened 2021-12-09
fix(Sentry): use correct release

guidojw:
New errors on production are being coupled to the [wrong release](https://sentry.io/organizations/csvalpha/releases/amber-ui%408a161b19d172cf23b4fb93861af65c9e2eecfea1/?project=186017) on Sentry, most likely because of release name in `config/environment.js`, so this PR fixes that. We use commit SHAs for release names and those are always unique so the amber-ui@ prefix is not necessary.

Release they should be coupled to: https://sentry.io/organizations/csvalpha/releases/8a161b19d172cf23b4fb93861af65c9e2eecfea1/?project=186017

*Edit:* added two other fixes:

* When deploying to master with a merge, the commit SHA of the head commit before the merge was used in the release action.
* Removed environment from the release action call in Publish Image, to not have Sentry create a deployment ([reference](https://github.com/getsentry/action-release/blob/744e4b262278339b79fb39c8922efcae71e98e39/src/main.ts#L56)).


DrumsnChocolate:
I assume you mean we can leave out the project name prefix in the release settings, because we only use sentry for this project?

guidojw:
> I don't understand much of the contents of the yaml files, but the changes are fairly small and seem reasonable to me. Have you been able to verify that this does indeed work, or are you hoping this fixes the issue once pushed to production?

Yes, similar action configurations works on my own projects.

DrumsnChocolate (approved):

PR #387, opened 2021-12-09
Webauthn MFA

Matthijsy:
### Summary
See https://github.com/csvalpha/amber-api/pull/267


wilco375:
Closing due to inactivity and relatively large codebase increase with code almost nobody will use.

PR #412, opened 2021-12-13
feat: new Dockerfile base image

guidojw:
### Summary
I originally put the ember-cli dependencies in de Ember.js group in #372, but found out that ember-cli versions are independent of Ember.js updates and newer CLI versions can still build older Ember.js apps. This PR changes the base image in the Dockerfile to another ember-cli image which has the following pros and cons:

**Pros**

- It's updated more frequently
- Its Node.js version actually matches most up-to-date [ember-cli support](https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md) (madnificent's (one we use now) doesn't sadly)
- Unblocks #399 (due to compatible Node.js version, above point)
- It looks like this can also easily be used in a development environment (I tried in #235, did not succeed, can try again with this one as base in the future)
- It installs Chrome drivers so makes `Dockerfile.test` redundant but the overall size is a bit bigger, the bundle size isn't though since the NGINX stage in the Dockerfile still does the same

**Cons**

- The overall size is a bit bigger

I also used [digest pinning](https://docs.renovatebot.com/docker/#digest-pinning) for the Docker image versions of ember-cli and NGINX.

### Other information

- https://github.com/danlynn/ember-cli


guidojw:
Tests fail because this needs #411

PR #423, opened 2022-01-19
fix: use variable instead of value as default for BUILD_HASH

guidojw:
oops


DrumsnChocolate (approved):

DrumsnChocolate:
"Did you test it???"
"Yes yes I tested it"

guidojw:
image

PR #445, opened 2022-02-01
Fix localization of moment dates

wilco375:
### Summary
Dates were always shown in English, regardless of locale. This PR fixes that.


guidojw:
Maybe this fits better in a specific initializer for moment?

PR #449, opened 2022-02-03
PWA

wilco375:
### Summary
This PR makes the Alpha site a Progressive Web App. This way, you should be able to install it as a 'normal' app, and it will also show up as such in the app drawer / home screen. AFAIK there is no way to test if this works properly right now, as one of the requirements for a PWA is the usage of HTTPS, which is not available in a dev environment, but we can see if it works on staging 😄 

What I did verify is that the offline page is working (once you shut down the ember server it shows the offline page). Furthermore, I inspected the application tab of the Chrome DevTools, which did not show any errors after my changes.

image


guidojw:
I have been using it as an app for a while, both on my previous Android and current iPhone
image

Is it any different than this?

wilco375:
@guidojw 
I don't know about iPhone, but for Android it shows up as an installed application in the app drawer, instead of just as a shortcut on the home screen. Furthermore, I have added some shortcuts that, on Android, can be opened when long pressing the application icon. This is also only possible for PWAs and not for shortcuts.

guidojw:
The screenshots above also aren't just shortcuts because they open in a special window, not in Safari (it worked like this on my Android too).

My main point is: the serviceworker stuff may be somewhat unnecessary as it already works correctly, the shortcuts in the manifest.json are a nice addition though. 
Also is the offline page necessary? We currently have that notification "the API is currently unavailable" right, and I think browsers keep the assets cached for a while.

Here's probably why it currently already kinda works:
https://github.com/csvalpha/amber-ui/blob/ce4224b9ad9ea7be9548dd6aec713111ff686872/app/index.html#L12

wilco375:
Some screenshots to show the difference:

**Current Alpha site, with shortcut**
Screenshot_20220203-152406

**Example website using PWA (install app instead of add shortcut)**
Screenshot_20220203-152357

It is then actually installed as a system app and behaves as such, instead of a shortcut:
Screenshot_20220203-152509
Screenshot_20220203-153113-375

As I said, the PWA shortcuts are also an advantage, see: https://web.dev/app-shortcuts/

wilco375:
> 
> Here's probably why it currently already kinda works (+ the comment & tag a few lines down):
> 
> https://github.com/csvalpha/amber-ui/blob/ce4224b9ad9ea7be9548dd6aec713111ff686872/app/index.html#L12

Also, this is probably why it works like that on iOS but on Android it works differently

wilco375:
BTW, fun bonus, PWAs can also be installed as apps on desktop: 
image

guidojw:
Okay looks good, I was convinced that it "opening in another window" meant it was some kind of PWA, but that seems to not be true, apologies.

Then only my point about the offline page still stands

tests/rendering/components/md-editor-test.js
```diff
@@ -7,6 +7,7 @@ module('Integration | Component | markdown editor', function(hooks) {
   setupRenderingTest(hooks);
 
   test('should render the textarea content', async function(assert) {
+    debugger;
```
guidojw, line 10: Is this meant to be here?

public/offline.html
```diff
@@ -0,0 +1,32 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport"
+        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <title>DARF - Offline</title>
```
guidojw, line 8: What is DARF?

tests/rendering/components/md-editor-test.js
wilco375: Oops no

public/offline.html
wilco375: Oops, copied this from another project...

wilco375:
> Okay looks good, I was convinced that it "opening in another window" meant it was some kind of PWA, but that seems to not be true, apologies.
> 
> Then only my point about the offline page still stands

I guess I can remove it, yeah

guidojw (approved):

DrumsnChocolate:
Hmm, I also thought this was already a thing, but apparently not.

PR #455, opened 2022-02-07
feat: show numeric keyboard for OTP code

guidojw:
### Summary
Adds the pattern attribute to the auth code input so that iOS devices will show a numeric keyboard instead (Android does not do this).


wilco375:
Can't we just make it a type="number" field?

guidojw:
No, that won't force a numeric keyboard on iOS and will show increment/decrement buttons which do not make much sense for this field

wilco375:
You should be able to add `inputmode="numeric"` for Android. Can you also add that?
https://css-tricks.com/finger-friendly-numerical-inputs-with-inputmode/#aa-a-better-solution-inputmode

guidojw:
That seems useful too, added 👍🏿

wilco375 (approved):

PR #460, opened 2022-02-08
Fix relative path, update shortcuts

wilco375:
### Summary
This PR has two fixes:
1. Make the relative path to sw.js an absolute path so it can also be resolved on pages other than the home route
2. Update the shortcuts to most used pages according to Google Analytics:
afbeelding


public/manifest.json
```diff
@@ -48,20 +48,20 @@
...
       "icons": [{ "src": "/images/shortcuts/forum.png", "sizes": "192x192" }]
     },
     {
-      "name": "Polls",
-      "url": "/polls",
-      "icons": [{ "src": "/images/shortcuts/polls.png", "sizes": "192x192" }]
+      "name": "Groepen",
+      "url": "/groups",
+      "icons": [{ "src": "/images/shortcuts/groups.png", "sizes": "192x192" }]
     },
```
guidojw, line 65:
```suggestion
    }
```

guidojw (approved):

PR #470, opened 2022-02-10
feat: prettier

guidojw:
### Summary
Here's the ginormous prettier PR I talked about yesterday.
This adds the exact prettier configuration from the ember-cli [app blueprint](https://github.com/ember-cli/ember-new-output), that otherwise would have been added by [ember-cli-update](https://github.com/csvalpha/amber-ui/pull/469) soon (we can opt-out of course but I think the pros outweigh the cons). I then ran `prettier --write .` to format all the files prettier possibly can format and fixed all the ESLint and Stylelint errors.

`ember-cli-template-lint` is deprecated so I changed it to `ember-template-lint` (its successor).

### Other information
* The relevant RFC: https://emberjs.github.io/rfcs/0628-prettier.html
* You can run prettier on save in WebStorm: https://blog.jetbrains.com/webstorm/2020/02/webstorm-2020-1-eap-5/ (I set "Run for files" to `{**/*,*}.*` since prettier also formats hbs, scss, md, ...)


wilco375 (approved):
Haven't looked through every file, but everything seems to be working

Issue #424, opened 2022-01-19
Feature: swiping in photo albums


DrumsnChocolate:
### How to reproduce
Go to a foto album, click on a photo

### Desired behaviour
Swiping left and right on mobile should work similarly to pressing left and right arrow keys on desktop, redirecting to next or previous image in album.

### Current behaviour
Swiping doesn't do a thing

guidojw:
https://github.com/Ember-Swiper/ember-cli-swiper this looks cool, doesn't look too maintained though.

DrumsnChocolate:
I don't think it needs animation, so it might be simpler to use touch events on HTML. We should be able to write a generic Swipable component using touchstart, touchmove, touched and touchcancel, I think: https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent

PR #471, opened 2022-02-11
Feature: swipable photos

DrumsnChocolate:
implements #424


wilco375 (dismissed):
Can you remove the console logs? Other than that it looks good and seems to be working fine.

app/components/swipeable.js
```diff
@@ -0,0 +1,83 @@
...
+    this.yStart = null;
+    this.xEnd = null;
+    this.yEnd = null;
+  }
+
+  @action
+  onTouchStart(touchStartEvent) {
+    if (touchStartEvent.targetTouches.length === 1) {
+      this.touching = true;
+      let touch = touchStartEvent.targetTouches.item(0);
```
guidojw, line 28: Can you use `const` instead of `let` on places where the variables do not change?

guidojw (changes requested):

app/components/swipeable.js
DrumsnChocolate: Sure, but shouldn't small things like these be caught by the linter?

DrumsnChocolate:
Lol, I just discovered the photo album swipes the wrong way. Tested it on an album with 2 photos first, now with 3, I see that it loops in the opposite direction of what the user would expect.

PR #475, opened 2022-02-13
Add data: to img-src in CSP

wilco375:
### Summary
This PR adds `data:` to the list of allowed `img-src`s in the CSP such that the base64 data images in e.g. the article cover upload preview can be shown. Since this only allows inline data in img context this should pose no inherent security risk.

Before fix:
image

After fix it previews the uploaded image.


guidojw (approved):
For this to work, you have to change the contentSecurityPolicy field in `traefik/dynamic/middlewares.yml` on the server to the output of `ember csp-headers --environment production`

PR #476, opened 2022-02-13
Fix duplicate tests

wilco375:
### Summary
Maybe I'm missing something, but there seem to be some duplicate tests that don't add anything. I have de-duplicated those. It might be the result of some merge.


guidojw (approved):
Found out about this too in #474 😋

PR #477, opened 2022-02-14
Feature/library

### LINKED ISSUE #446: Feature idea: Alpha Library
In the board room we have a library of books that members can use for for example _kring_. It is to my knowledge not really used currently, probably partly because nobody knows what's there. It would be nice if there was a page on the website with all the available books.

wilco375:
### Summary
Fixes #446 
Also see https://github.com/csvalpha/amber-api/pull/278

Features:
- Create, view, update, delete books
- Barcode scanner using camera to scan ISBN numbers
- ISBN lookup on bol.com partner API for automatically filling in fields

image
image
image


app/controllers/books/edit.js
```diff
@@ -0,0 +1,72 @@
...
+    book.set('coverPhoto', file.data);
+  }
+
+  @action
+  submit() {
+    if (!this.model.isbn) {
+      this.model.isbn = null;
+    }
+
+    this.model.save().then(() => {
```
guidojw, line 28: I think if you do `super.submit()` here the EditController's submit method will do exactly the same stuff

app/controllers/books/new.js
```diff
@@ -0,0 +1,3 @@
+import EditController from './edit';
+
+export default EditController;
```
guidojw, line 3: Ember doesn't like when you re-export components, for controllers it may work but maybe it's nicer and clearer intention when you do `export default class NewBookController extends EditController {}`?

app/routes/books/destroy.js
```diff
@@ -0,0 +1,13 @@
+import { AuthenticatedRoute } from 'alpha-amber/routes/application/application';
+
+export default class DestroyGroupRoute extends AuthenticatedRoute {
```
guidojw, line 3: Book*

app/models/book.js
```diff
@@ -0,0 +1,25 @@
+import Model, { attr } from '@ember-data/model';
+
+export default class Article extends Model {
```
guidojw, line 3: Book*

guidojw (changes requested):

app/templates/books/show.hbs
```diff
@@ -0,0 +1,29 @@
+<EmberWormhole @to='navbar-wormhole'>
+  <PageActionsButtons @pageActions={{pageActions}} />
+</EmberWormhole>
+
+<div class='card' itemscope itemtype='https://schema.org/Book'>
+  <div class='card-body p-3'>
+    <img
+      src='{{model.coverPhotoUrlOrDefault}}'
```
guidojw, line 8: Won't this make it a string?

guidojw (approved):

PR #498, opened 2022-03-14
feat: actionlint

guidojw:
### Summary

Runs [some checks](https://github.com/rhysd/actionlint/blob/main/docs/checks.md) on our workflow files.

### Other information

https://github.com/rhysd/actionlint


.github/workflows/continuous-delivery.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         env:
           INPUT_MERGE: ${{ github.event.inputs.merge }}
         run: |
-          if [ $GITHUB_REF_NAME = 'master' ]; then
+          if [ "$GITHUB_REF_NAME" = 'master' ]; then
             if [ "${INPUT_MERGE,,}" = 'y' ]; then
               git fetch origin staging
               if ! git diff origin/master origin/staging --exit-code; then
```
DrumsnChocolate, line 50: I know this isn't related to this commit, but I'm wondering whether this condition is not the wrong way around; doesn't a non-zero exit code mean that there's differences between branches? According to `man git-diff`:

>        --exit-code
>           Make the program exit with codes similar to diff(1). That is, it
>           exits with 1 if there were differences and 0 means no differences.
Wouldn't that then mean that a truthy exit code means differences, and passing the exit code through a negation results in true when no diff?
that is, shouldn't it be:
`if git diff --exit-code ... has_diff=true else has_diff=false`? rather than 
`if ! git diff --exit-code ... has_diff=true else has_diff=false`

DrumsnChocolate (approved):

.github/workflows/continuous-delivery.yml
guidojw: An exit code of 1 means the program has failed, and this negates that to a success and thus makes it pass the condition.

PR #500, opened 2022-03-15
feat: run ember-template-lint in CI

guidojw:
### Summary

Runs `ember-template-lint` in the Continuous Integration workflow, and fixes all (around 580) linting errors that existed. Did some other small style changes too.

Most of the errors were indentation errors, you would think prettier fixes that but it seems that prettier is not deterministic on hbs and passes multiple formats.


DrumsnChocolate:
can we get prettier to be deterministic?

guidojw:
I don't think so

DrumsnChocolate (approved):
checked that yarn lint didn't give errors, and ran prettier

guidojw:
> Did you perform auto fix using ember-template-lint? or some other way?

I did not even look for that lol, but when I run `ember-template-lint` on staging, it gives 569 errors, 0 fixable

DrumsnChocolate:
> > Did you perform auto fix using ember-template-lint? or some other way?
> 
> I did not look for that lol, but when I run `ember-template-lint` on staging, it gives 569 errors, 0 fixable

oh, when I ran `yarn lint:hbs` it gave me the same number of errors, but more than 400 of those were supposedly fixable. when I tried running `ember-template-lint --fix .` it gave an exception that I didn't understand and I gave up

PR #503, opened 2022-03-16
feat: simplify thread & photo navigation and deduplicate photo-comments implementation

DrumsnChocolate:
Since navigation can now be done through arrow keys and swiping for photos in an album, I added the same functionality to threads, and changed the way that the history is tracked in these environments. The previous route in history now no longer points to the previous photo in the album, or the previous page in the thread, but rather to the route that was visited before entering the thread or photo-album.
I also simplified the implementation of photo-comments, because comments are displayed and interacted with in the same way via the `photos.show` route as well as in the `photo-comments` route. I extracted separate components for displaying photos and photo-comments, and arranged them differently depending on the route based on some css remarks from @wilco375


DrumsnChocolate:
bump @guidojw @wilco375

guidojw (approved):
Looks good

PR #529, opened 2022-05-05
refactor: Glimmer Components

guidojw:
### Summary

Glimmer Components

Resolves #530

Issue #429, opened 2022-01-19
Ember Octane update list


guidojw:
Here is a list with everything that we have to update in order to be fully using Ember Octane:

- [ ] [Glimmer Components](https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/)
- [ ] ...

PR #539, opened 2022-05-10
refactor(model-form): Glimmer Components

guidojw:
### Summary

Realised in #529 that the model-form components need to be refactored quite a lot so thought it would be a good idea to do this in a separate PR.

### Other information

`yarn install` breaks the types currently, in a future release of `webpack` where they moved `@types/eslint-scope` to the `devDependencies`, this will be fixed.


DrumsnChocolate:
@guidojw what do you mean by yarn install being broken? is that why the build fails?

guidojw:
> @guidojw what do you mean by yarn install being broken? is that why the build fails?

It installs two versions of the same `@types` package and thus TypeScript building fails because of overlapping types. I fixed `yarn.lock` to have it not do that which works as a temporary fix in dev.
However because of the `--immutable` flag and yarn wanting to update `yarn.lock` because it's technically incorrect, the command in CI here fails.

https://github.com/webpack/webpack/pull/15772 should fix this.

(also this PR is far from finished)

DrumsnChocolate:
I did run into some build errors yeah, I just tried testing this PR and it tells me exactly what you mentioned, duplicate declarations of the same identifiers. I'll wait with reviewing this until you think it's ready then.

guidojw:
This now builds

DrumsnChocolate:
I think I will start on this PR once #635 and  #896 are merged

DrumsnChocolate:
> This is also the PR that introduces TypeScript via ember-cli-typescript, because I was gonna do the new components in TypeScript.

If you don't mind, I will separate this from the refactor to glimmer. I like the idea of using typescript for our components, but for the sake of getting up to date with other dependencies we need to upgrade our ember version asap. In contrast, TS is more of a nice to have.

Adding the introduction of TS to #429

PR #545, opened 2022-05-15
Fix swiping sensitivity

### LINKED ISSUE #543: Forum topic switches pages on scroll
When on a phone, a forum topic switches pages when scrolling down

DrumsnChocolate:
fix #543 
well, it isn't necessarily the perfect solution, because it's a UX thing, but this should go a long way in making the detection of swiping gestures less trigger happy.
Also tried to make the component logic a bit more readable and less verbose

### DIFF
--- a/app/components/swipeable.js
+++ b/app/components/swipeable.js
@@ -8,11 +8,10 @@ export default class SwipeableComponent extends Component {
   yStart = null;
   xEnd = null;
   yEnd = null;
-  swipeDirection = this.args.onSwipeHorizontal
-    ? 'horizontal'
-    : this.args.onSwipeVertical
-    ? 'vertical'
-    : null;
+  horizontally = !!this.args.onSwipeHorizontal;
+  vertically = !!this.args.onSwipeVertical;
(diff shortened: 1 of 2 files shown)


app/components/swipeable.js
```diff
@@ -72,17 +81,24 @@ export default class SwipeableComponent extends Component {
...
-        throw new Error('No swipe direction defined');
+        this.throwNoDirection();
       }
+      direction = this.sign(direction);
+      this.onSwipe(direction);
     }
   }
+
+  sign(n) {
+    return n > 0 ? 1 : -1;
```
guidojw, line 98: Can't you use [`Math.sign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign) here? This can return 0 though when you pass 0 into it but that's correct in case the touch ends at exactly the same location as it started I think

DrumsnChocolate: good point!

wilco375 (approved):

PR #553, opened 2022-05-18
Fix/proper ordering of closed question answers in form responses

### LINKED ISSUE #303: Activity registration selected options are not ordered
### How to reproduce
1. Create a activity with a "meerdere opties mogelijk" question. 
2. Submit a registration for the activity

### Expected behaviour
The options selected should be ordered

### Current behaviour
The selected options are not in order

Ellen-Wittingen:
fix #303


app/models/form/response.js
```diff
@@ -32,6 +32,11 @@ export default class Response extends DirtySaveModel {
             await this.closedQuestionAnswers
           ).map((closedQuestionAnswer) => closedQuestionAnswer.option)
         );
+        answers.sort((answer1, answer2) =>
+          answer1.option && answer2.option
+            ? answer1.option.get('position') - answer2.option.get('position')
+            : 0
+        );
```
DrumsnChocolate, line 39:
```suggestion
        answers.sortBy('position');
```

DrumsnChocolate (changes requested):

app/models/form/response.js
Ellen-Wittingen: This change does not show the correct ordering in the sign-up table rows when I try it

DrumsnChocolate: yeah I don't know why I thought this would work, and I'm unsure why I didn't see any change in the ordering when testing it.

DrumsnChocolate (approved):

PR #557, opened 2022-05-18
Fix md-editor not updating new text formatting

### LINKED ISSUE #308: Formatting text does not work
### How to reproduce
1. Go to a place where you can write some text with formatting (forum post for example)
2. Try to insert some formatting (for example bold, hyperlink)

### Expected behaviour
Text will appear that will create the formatting

### Current behaviour
Nothing happens

Ellen-Wittingen:
fix #308

### DIFF
--- a/app/components/md-editor.js
+++ b/app/components/md-editor.js
@@ -6,7 +6,7 @@ export default class MdEditorComponent extends Component {
   classNames = ['md-editor-and-toolbar'];
   editMode = true;
 
-  content = '';
+  @tracked content = '';
   @tracked textareaId;
 
   @action


DrumsnChocolate (approved):

wilco375 (changes requested):
As mentioned in #308, simply adding tracked causes another issue. The content is now no longer saved.

DrumsnChocolate:
> As mentioned in #308, simply adding tracked causes another issue. The content is now no longer saved.

Oops, I should have remembered that.
@Ellen-Wittingen did you check the comments in the issue before working on a fix?

DrumsnChocolate:
@Ellen-Wittingen I'll leave updating this PR to you. I think updating this in a similar fashion to #567 should take care of the concerns that guido brought up

Ellen-Wittingen:
I changed the data flow of the textarea content a bit. It used to be a confusing two-way path where data would travel and be changed between the model content (post text, article text, etc.) and the md editor. Now it is one-way: the textarea gets the content from the model. When the md-editor want to change the content, it does it thorugh a function passed from the model, which changes it in the model, which trickles back down to the md-editor. For clarity: I followed this way of thinking: [https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_one-way-data-flow](https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_one-way-data-flow)

Ellen-Wittingen:
Oops I broke  #567 again, I will fix that later.

Ellen-Wittingen:
I fixed #567 again, it was just a merge issue

wilco375 (changes requested):
Although the formatting is applied to the text, something doesn't seem to be working properly. When I try to create an article it tells me the content is empty when I try to save, even though it isn't. Also, it would be nice if, after clicking a format button, the textarea is again focused so you can continue typing without having to click on the textarea again first. Don't know if this is possible / easy to implement?
image

app/components/md-editor.js
```diff
@@ -1,17 +1,13 @@
-import Component from '@ember/component';
+import Component from '@glimmer/component';
 import { action } from '@ember/object';
-import { tracked } from '@glimmer/tracking';
 
 export default class MdEditorComponent extends Component {
   classNames = ['md-editor-and-toolbar'];
```
wilco375, line 5: This does not seem to work properly. The class is not added, causing there to be no line around the element.
Current:
image
New:
image

wilco375 (changes requested):

wilco375:
It would be nice if, after clicking a format button, the textarea is again focused so you can continue typing without having to click on the textarea again first. Don't know if this is possible / easy to implement?

DrumsnChocolate (commented):
changes are looking nicely simple, no unnecessarily complex patterns to be found. Applying style options does seem to work, but there are some minor things that still need to be taken care of, see comments regarding previewing and the `classNames` API

wilco375 (approved):

DrumsnChocolate (approved):
nice work!

DrumsnChocolate:
@Ellen-Wittingen merge?

PR #567, opened 2022-05-31
Fix new forum post content resetting

### LINKED ISSUE #520: bug: text in the reageren box on the forum is not reset when visiting a different thread
### How to reproduce

1. Write something in the reageren box in a forum thread.
2. Navigate to a different forum thread using any buttons on the website (so don´t refresh the page or type the url yourself).
3. Look at the reageren box in that thread.

### Expected behaviour

When you look at the second visited forum thread, the reageren box should contain no text, as a user probably does not want to reuse their reaction text for the first thread in the second thread.

### Current behaviour

The text that was in the reageren box in the first thread is carried over to the reageren box in the second thread

### LINKED ISSUE #641: forum-post-new component does not clear contents upon cancel
brought up in https://github.com/csvalpha/amber-ui/pull/556#discussion_r996296640

Ellen-Wittingen:
fix #520 and fixes that the cancel button on the new forum post actually clears the content
fix #641


app/components/forum/forum-post-new.js
```diff
@@ -25,6 +25,6 @@ export default class ForumPostNewComponent extends Component {
 
   @action
   cancel() {
-    this.content = '';
+    this.set('content', '');
```
guidojw, line 28: We have to consider that this is not "Octane style" and will give a lint error once https://github.com/csvalpha/amber-ui/blob/fca981602f54d818300d07c4b346b347b6c34f1d/.eslintrc.js#L30 is reenabled.

DrumsnChocolate:
I've turned the ForumPostNew component into a glimmer component. I'm not sure why things were broken in the classic component, but I also don't want to bother with that, since this new implementation fixed the problem.
For reference, data binding is now one-way (that's a trait of glimmer components): https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_one-way-data-flow
This means the new component now modifies contents by calling a function passed in its arguments.

app/routes/forum/categories/category/threads/thread/posts/index.js
```diff
@@ -68,6 +68,24 @@ export default class PostIndexRoute extends AuthenticatedRoute {
...
+          this.modelFor('forum.categories.category.threads.thread')?.id
+      )
+    ) {
+      const controller = this.controllerFor(routeName);
+      if (controller) {
+        controller.send('resetNewContent');
+      }
+    }
+  }
+
```
DrumsnChocolate, line 88: haven't touched this, will have a look at whether this is octane style

app/routes/forum/categories/category/threads/thread/posts/index.js
```diff
@@ -68,6 +68,24 @@ export default class PostIndexRoute extends AuthenticatedRoute {
     this.controller.send('goToNextPage');
   }
 
+  @action
+  beforeModel(transition) {
+    const routeName = 'forum.categories.category.threads.thread.posts.index';
```
DrumsnChocolate, line 73: this is unnecessary. we can just use `this.routeName`

app/routes/forum/categories/category/threads/thread/posts/index.js
DrumsnChocolate: I've simplified this.
see https://api.emberjs.com/ember/4.8/classes/Route/methods/resetController?anchor=resetController

app/routes/forum/categories/category/threads/thread/posts/index.js
```diff
@@ -68,6 +68,13 @@ export default class PostIndexRoute extends AuthenticatedRoute {
     this.controller.send('goToNextPage');
   }
 
+  resetController(controller, isExiting, transition) {
```
DrumsnChocolate, line 71: https://api.emberjs.com/ember/4.8/classes/Route/methods/resetController?anchor=resetController for docs on this hook

DrumsnChocolate (approved):
it's a bit weird to approve my own changes

app/components/forum/forum-post-new.js
```diff
@@ -9,20 +9,19 @@ export default class ForumPostNewComponent extends Component {
 
   @action
   async save() {
-    let { content, thread } = this;
```
wilco375, line 12: You could also update this line to `= this.args` instead of doing that in `createRecord` below; not sure what's more readable

wilco375 (approved):

DrumsnChocolate:
see #657 for a silly extra PR that fixes the same thing. I forgot this PR existed, but apparently I solved it in a very similar manner. I added some small stuff added in #657 by merging the PR directly into this branch. See the comments on that PR for a detailed explanation.
> You could also update this line to = this.args instead of doing that in createRecord below; not sure what's more readable

This is now also taken care of.

app/controllers/forum/categories/category/threads/thread/posts/index.js
```diff
@@ -55,8 +54,12 @@ export default class PostsIndexController extends Controller {
 
   @action
   async addQuote(q) {
-    this.set('newContent', `${this.newContent}${q} \n\n`);
-
+    if (!this.newContent?.endsWith('\n')) {
```
guidojw, line 57: Unnecessary chaining :p, confused me because now it looks like `this.newContent` could be undefined/null and this conditional will pass then because !undefined/!null == true

PR #589, opened 2022-08-04
Fix code in file-upload.hbs using deprecated components

### LINKED ISSUE #587: New activity form only shows up half so users cannot make new activity
The new activity form only shows up half, see the image for the current issue. Because of this users cannot make new activities.
Screenshot_20220803_154419

I tracked it down to the module ember-file-upload being updated, which deprecated the `FileUpload` tag, which is used in `app/components/model-form/file-input.hbs` which is used for the new activity form. Their documentation shows a way to fix it under the sections "Upgrading to v5" Looking at that, I changed 
```
<FileUpload
    @name='fileUpload'
    @accept={{validMimetypes}}
    @onFileAdd={{action 'fileLoaded'}}
    @class='clickable d-block file-picker'
  >
    {{#if (get model property)}}
      <img
        id={{inputIdentifier}}
        src={{mut (get model property)}}
        alt={{label}}
      />
      <br />
    {{else}}
      <span class='image-upload-text'>
        Sleep hier een nieuwe bestand naartoe of klik om een bestand te kiezen
      </span>
    {{/if}}
  </FileUpload>
```
to
```
{{#let (file-queue name="fileUpload" onFileAdded=actions.fileLoaded) as |queue|}}
    <label for="upload-file">
      {{#if (get model property)}}
        <img id={{inputIdentifier}} src={{mut (get model property)}} alt={{label}} role="button" tabindex="0"
          aria-controls="upload-file" />
        <br />
      {{else}}
        <span class='image-upload-text' role="button" tabindex="0" aria-controls="upload-file">
          Sleep hier een nieuwe bestand naartoe of klik om een bestand te kiezen
        </span>
      {{/if}}
    </label>
    <input type="file" id="upload-file" class="" hidden accept={{validMimetypes}} {{queue.selectFile}}>
  {{/let}}
```
However, now there is a new issue where suddenly the function `this.set()` does not seem to be recognised (see image).
Screenshot_20220803_154246

I tried to import `set` but that did not work. Then I tried to make the variables `fileTooBig` and `extensionNotPermitted` tracked to update them without `set()` but then I got other errors. I think this issue is caused by the way `onFileAdded=actions.fileLoaded` calls the action, but I do not know how to tell it to call the action `fileLoaded` with `this` properly linked. So that still needs to get fixed.

Ellen-Wittingen:
fix #587

### DIFF
--- a/app/components/model-form/file-input.hbs
+++ b/app/components/model-form/file-input.hbs
@@ -2,11 +2,11 @@
   {{label}}{{if required ' *'}}
 </label>
 <div class='{{inputWrapperClass}}'>
-    {{#let (file-queue name="fileUpload" onFileAdded=(action 'fileLoaded')) as |queue|}}
+  {{#let (file-queue name="fileUpload" onFileAdded=(action 'fileLoaded')) as |queue|}}
     <label for="upload-file">
       {{#if (get model property)}}
         <img id={{inputIdentifier}} src={{mut (get model property)}} alt={{label}} role="button" tabindex="0"
-          aria-controls="upload-file" />
+        aria-controls="upload-file" />
         <br />
       {{else}}
         <span class='image-upload-text' role="button" tabindex="0" aria-controls="upload-file">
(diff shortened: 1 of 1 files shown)


guidojw (approved):
Nice!

PR #646, opened 2022-10-30
Fix publish-image workflow

wilco375:
Fixes newlines in publish-image workflow
See https://github.com/orgs/community/discussions/26288#discussioncomment-3876281 for reference


.github/workflows/publish-image.yml
```diff
@@ -41,12 +41,14 @@ jobs:
...
+            echo "build_args<<${delimiter}" >> "$GITHUB_OUTPUT"
+            BUILD_ARGS="BUILD_HASH="${INPUT_SHA:-$GITHUB_SHA}
             if [ "$GITHUB_REF_NAME" = 'staging' ]; then
-              BUILD_ARGS+=$'\nDEPLOY_TARGET=staging'
+              BUILD_ARGS+="\nDEPLOY_TARGET=staging"
             fi
-            BUILD_ARGS=${BUILD_ARGS//$'\n'/'%0A'}
-            echo 'build_args='"$BUILD_ARGS" >> "$GITHUB_OUTPUT"
+            echo $BUILD_ARGS >> "$GITHUB_OUTPUT"
+            echo "${delimiter}" >> "$GITHUB_OUTPUT"
```
guidojw, line 51:
```suggestion
            BUILD_ARGS='BUILD_HASH='${INPUT_SHA:-$GITHUB_SHA}
            if [ "$GITHUB_REF_NAME" = 'staging' ]; then
              BUILD_ARGS+=$'\nDEPLOY_TARGET=staging'
            fi
            echo 'build_args<<EOF' >> "$GITHUB_OUTPUT"
            echo "$BUILD_ARGS" >> "$GITHUB_OUTPUT"
            echo 'EOF' >> "$GITHUB_OUTPUT"
```
* Moved the start echo of build_args to where the rest is outputted to logically group building/setting output together.
* Removed openssl stuff as it overcomplexifies it and isn't necessary here as commit shas cannot possibly contain EOF because they're fully lowercase/numbers and we're not working with user input here.
* Readded $'' strings, normal strings do not handle newline characters correctly (at least single quote strings, do double quote strings?) and they are essential here.
* Changed strings that do not need expansion back to single quote, as was done before and elsewhere too.

guidojw (changes requested):

.github/workflows/publish-image.yml
wilco375: I think there is supposed to be a literal \n in the text now. It is also that way in the example mentioned above

wilco375: Nevermind, I think you're right. They might be using a different shell than bash

guidojw (approved):

PR #655, opened 2022-11-19
Update Flux contact info in footer

wilco375:
### Summary

Often, the phone number of Flux is called when somebody wants to contact Alpha. This PR replaces the contact info of flux with a link to the website of Flux, so that the differentiation Alpha<>Flux is hopefully a bit clearer.

### DIFF
--- a/app/components/footer-nav.hbs
+++ b/app/components/footer-nav.hbs
@@ -112,46 +112,6 @@
               <tbody>
                 <tr>
                   <td class='footer-table-cell'>
-                    <FaIcon @icon='envelope' />
-                  </td>
-                  <td class='footer-table-cell'>
-                    <a
-                      class='footer-text'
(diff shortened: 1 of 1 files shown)


app/components/footer-nav.hbs
```diff
@@ -108,77 +108,52 @@
             <h5 class='footer-header' itemprop='name'>
               Stichting Sociëteit Flux
             </h5>
-            <table class='table table-sm'>
+            <table class='table table-sm mw-100'>
```
wilco375, line 111: By default, the `max-width` of table is `inherit`, which means it will take the `max-width` of the parent, on desktop being `33%` (width of `.col-md-4`). This causes the table to wrap text to new lines unnecessarily, most notably the address.

DrumsnChocolate (approved):

PR #651, opened 2022-11-06
refactor: remove `ember-crumbly`

guidojw:
### Summary

Replaces `ember-crumbly` with a custom implementation based on https://github.com/poteto/ember-crumbly/issues/149.

Had to restructure the routes in order to retain the parent route breadcrumbs. Due to this restructure, the `show` routes could be renamed to `index` routes, making more sense. Also after our conversation on Slack it looked like the forum routes were working correctly as is, but this wasn't the case so these routes needed some restructuring as well, overall unifying the route & controller structure.

I noticed there was an unimplemented `profile` route being created in router.js, so I implemented this to redirect to your own user page. `users/batch/confirm` route also was unimplemented so I removed it.

Lastly, I restructured the tests to the new structure too and implemented all missing route unit tests.

Todo:
- [x] PR to amber-api where `forgot_password` and `activate_account` in the activation mail are updated to `forgot-password` and `activate-account`. https://github.com/csvalpha/amber-api/pull/353

PR #656, opened 2022-11-19
Refactor: remaining controllers to octane

### LINKED ISSUE #618: Not all controllers are in octane style
aside from #555 , there are other controllers that still need to be converted to octane

DrumsnChocolate:
fix #618

### DIFF
--- a/app/controllers/mail-moderations/reject.js
+++ b/app/controllers/mail-moderations/reject.js
@@ -1,6 +1,8 @@
 import { inject as service } from '@ember/service';
 import Controller from '@ember/controller';
 import { isInvalidResponse } from 'ember-fetch/errors';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
 // todo: flash notice?
 export default class RejectController extends Controller {
   @service fetch;
--- a/app/controllers/mail-moderations/reject.js
+++ b/app/controllers/mail-moderations/reject.js
@@ -1,24 +1,23 @@
 import { inject as service } from '@ember/service';
 import Controller from '@ember/controller';
 import { isInvalidResponse } from 'ember-fetch/errors';
-// todo: octane
-export default Controller.extend({
-  fetch: service(),
-  errorMessage: null,
-  actions: {
-    async submit() {
-      this.set('errorMessage', null);
-      const response = await this.fetch.post(
-        `/stored_mails/${this.model.id}/reject`
-      );
+// todo: flash notice?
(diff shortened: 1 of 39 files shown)


wilco375 (approved):
Nicely done! I've tested the relevant pages and have not come across any issues

app/controllers/sog/name-trainer.js
```diff
@@ -19,131 +22,164 @@ export default Controller.extend({
...
+    currentQuestion.success =
+      currentQuestion.answer === currentQuestion.question.get('fullName');
+    if (currentQuestion.success) {
+      this.success++;
+    }
+    this.goToNextQuestion();
+  }
+
   generateQuestions(people) {
+    if (!people.length) {
```
guidojw, line 140: I don't like this control flow and think you should rather check people.length in the startTrainer action and set errorMessage there directly, as generating questions for no people should logically return no questions instead of an error

app/models/debit/collection.js
```diff
@@ -26,7 +26,9 @@ export default class Collection extends Model {
     });
     if (failedTransactions) {
       const prefix = failedTransactions > 1 ? 'zijn' : 'is';
-      throw `Er ${prefix} ${failedTransactions} transacties niet juist opgeslagen`;
+      throw new Error(
```
guidojw, line 29: 🚀

guidojw (changes requested):
Looking good 👍🏿, can't wait to resolve conflicts on #651

app/controllers/sog/name-trainer.js
DrumsnChocolate: good point, I think was a bit tired when I wrote this.

app/controllers/mail-moderations/reject.js
```diff
@@ -1,24 +1,25 @@
...
-  errorMessage: null,
-  actions: {
-    async submit() {
-      this.set('errorMessage', null);
-      const response = await this.fetch.post(
-        `/stored_mails/${this.model.id}/reject`
-      );
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+// todo: flash notice?
```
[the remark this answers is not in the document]
DrumsnChocolate: I kinda forgot

app/controllers/sog/name-trainer.js
DrumsnChocolate: fixed, tell me if this is what you had in mind

guidojw (approved):

PR #657, opened 2022-11-19
Ameliorate forum-post-new reactivity

### LINKED ISSUE #641: forum-post-new component does not clear contents upon cancel

DrumsnChocolate:
ref #641 
rewrote to glimmer component, because I understand them better, and it allows me to use the `this.args` namespace (see https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_benefits-of-glimmer-components)
note that args are immutable in glimmer components, which requires an extra function to be passed as an argument. Via this function `onChangeContent` we can pass the new value up to the containing controller, in this case `posts/index`
(If you're familiar with Vue.js then this pattern should look familiar.)

This would be fixed when #539 is finished, but I thought it wouldn't hurt to fix this since creating forum post is still broken right now

### DIFF
--- a/app/components/forum/forum-post-new.hbs
+++ b/app/components/forum/forum-post-new.hbs
@@ -1,5 +1,5 @@
 <p class='card-text'>
-  <MdEditor @content={{content}} @textareaId='newForumPost' />
+  <MdEditor @content={{this.content}} @textareaId='newForumPost' />
 </p>
 <div class='card-footer'>
   <div class='row justify-content-end'>
--- a/app/components/forum/forum-post-new.js
+++ b/app/components/forum/forum-post-new.js
@@ -1,5 +1,5 @@
 import { inject as service } from '@ember/service';
-import Component from '@ember/component';
+import Component from '@glimmer/component';
 import { action } from '@ember/object';
 
 // todo: incorporate the model-save-util into components?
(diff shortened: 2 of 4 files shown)


app/components/forum/forum-post-new.hbs
```diff
@@ -1,5 +1,5 @@
 <p class='card-text'>
-  <MdEditor @content={{content}} @textareaId='newForumPost' />
+  <MdEditor @content={{this.content}} @textareaId='newForumPost' />
```
DrumsnChocolate, line 2: note that this no longer refers to the `content` argument, but rather to the `content` value that comes from the controller. In this situation, that `content` property on the controller comes in the form of a getter.
If we wanted to still refer to the `content` argument, we would need to write `@content`
(this is an effect of glimmer components)
I hope this demonstrates how glimmer components clarify what we are referencing

DrumsnChocolate: I also find that this nicely demonstrates how glimmer components will clarify code in general

DrumsnChocolate:
wait, this is already being taken care of in #657

guidojw:
Ameliowhat

DrumsnChocolate:
> Ameliowhat

https://www.merriam-webster.com/dictionary/ameliorate
image

wilco375:
> Ameliowhat

Haha exactly my thought when I read this PR

PR #669, opened 2022-12-04
Fix time offset by one day

### LINKED ISSUE #667: Date one day off
### How to reproduce

1. Go to an edit page with a date field
2. Enter a date 

### Expected behaviour

The entered date is saved

### Current behaviour

Sometimes, the saved date is off by one day

wilco375:
### Summary
Fixes #667

### DIFF
--- a/app/components/input/date-input.js
+++ b/app/components/input/date-input.js
@@ -6,6 +6,7 @@ import { inject as service } from '@ember/service';
 
 export default TimeInputComponent.extend({
   media: service(),
+  type: 'date',
   momentFieldsToUpdate: ['year', 'month', 'date'],
   showNativeDateInput: oneWay('media.isMobile'),
 
--- a/app/components/input/date-input.js
+++ b/app/components/input/date-input.js
@@ -28,7 +28,8 @@ export default TimeInputComponent.extend({
 
   actions: {
     onPikadaySelection(selectedDate) {
-      this.updateDateValue(moment(selectedDate));
+      let date = moment(selectedDate);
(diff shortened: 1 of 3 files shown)


app/components/input/date-input.js
```diff
@@ -28,7 +28,8 @@ export default TimeInputComponent.extend({
 
   actions: {
     onPikadaySelection(selectedDate) {
-      this.updateDateValue(moment(selectedDate));
+      let date = moment(selectedDate);
+      this.updateDateValue(date.add(date.utcOffset(), 'minutes'));
```
wilco375, line 32: By adding the offset of the timezone to the date, we make sure that the UTC time is the same as the local time (e.g. 05-12-2022 00:00 UTC+2 becomes 05-12-2022 00:00 UTC instead of 04-12-2022 22:00 UTC)

guidojw: To me this seems like a fix to the wrong problem, but I may be wrong

We currently only do setLocale on moment in the application controller, can you try if .setTimeZone('Europe/Amsterdam') (found in ember-moment docs) fixes this?

DrumsnChocolate:
If I read #667 correctly, shouldn't we be fixing the way that dates fetched from the backend are displayed? rather than fixing the way they are input? Since the correct date was being saved?
Like, if we save 00:00 in local time, and this translates to 22:00 previous day in UTC, then that is correct, right? there's no need to fix the input logic there.
Can we try to find the underlying problem? because this feels like somewhat of a bandaid

wilco375:
Closing because of new fix: https://github.com/csvalpha/amber-ui/pull/682

PR #670, opened 2022-12-04
Add nicknames

wilco375:
### Summary
Adds a nickname field to the website.
Currently, some people have the nickname in brackets behind their first name. This PR will give a more robust solution for nicknames.

### Other information
See https://github.com/csvalpha/amber-api/pull/346 and https://github.com/csvalpha/sofia/pull/833

### DIFF
--- a/app/models/user.js
+++ b/app/models/user.js
@@ -1,4 +1,6 @@
+import Ember from 'ember';
 import { inject as service } from '@ember/service';
+import { htmlSafe } from '@ember/template';
 import Model, { hasMany, attr } from '@ember-data/model';
 import { isEmpty } from '@ember/utils';
 import moment from 'moment';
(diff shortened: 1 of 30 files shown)


app/models/user.js
```diff
@@ -71,6 +74,38 @@ export default class User extends Model {
...
+
+  get fullNameWithNickname() {
+    return this._fullNameWithNickname();
+  }
+
+  get fullNameWithNicknameMd() {
+    return this._fullNameWithNickname(true);
+  }
+
+  _fullNameWithNickname(markdown = false) {
```
Ellen-Wittingen, line 93: Nickname on SOFIA
image

Nickname on AMBER
image

On AMBER the nickname is put after the lastNamePrefix, but on SOFIA it is put before the lastNamePrefix. I think the one on SOFIA looks a lot better, so can you also put the nickname before the lastNamePrefix here? I believe you just have to sawp two if-statements in this method

Ellen-Wittingen (approved):

PR #676, opened 2022-12-12
Refactor components to octane

wilco375:
### Summary
This PR refactors classic Ember components to Octane-style components


guidojw:
For consistent naming: we also include Controller in controllers and Route in routes, so I think it's a good idea to include Component in the component class names.

PR #695, opened 2023-01-14
Fix quickpost

DrumsnChocolate:
fix button spacing, rewrite to glimmer, and change the sorting method
these are attempts at fixing the problems with the qp, they solve part of the problem but it seems that some issue with randomly reordering still remains


DrumsnChocolate:
Also made sure polling is working, and it turns out that a subscription is created to the channel `'/quickpost_messages'`, but the responses returned by `amber-api` to the polling requests don't seem to be always be complete

wilco375 (approved):

PR #716, opened 2023-03-08
fixed bounding of group photos

### LINKED ISSUE #615: Group pictures are not properly bounded
image
see picture. no vertical size limits
something that may help in fixing this: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

Macanazo:
fix #615

### DIFF
--- a/app/styles/routes/groups.scss
+++ b/app/styles/routes/groups.scss
@@ -1,4 +1,4 @@
 .groups-avatar {
   max-height: 75vh;
-  object-fit: contain !important;
+  object-fit: contain;
 }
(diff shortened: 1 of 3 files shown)


app/styles/routes/groups.scss
```diff
@@ -0,0 +1,4 @@
+.groups-avatar {
+  max-height: 75vh;
+  object-fit: contain !important;
```
wilco375, line 3: Is the !important needed here? Haven't tried it but I think this object-fit should overwrite the .card-img-rounded object-fit without !important, since this style file is imported later.

Macanazo: I have also not tried it without the !important so I would not know

Ellen-Wittingen (approved):

PR #737, opened 2023-04-03
Dies 2023

DrumsnChocolate:
based on https://github.com/csvalpha/alpha-banana-ui/pull/1376
tried to reassign the `$brand-primary` variable when authenticated, but that is harder than it seems at first glance.
@Ellen-Wittingen mentioned something about potentially reassigning all the generated css vars (with `--` prepended), but I don't have enough time to figure that out.
@wilco375 mentioned the option of changing `$brand-primary` altogether and overriding all usages in the external website, but I'm not too much of a fan of the work involved with that.

needs to be deployed on 21-04-2023


app/components/header-nav.hbs
```diff
@@ -32,7 +41,10 @@
 
     {{#unless session.isAuthenticated}}
       <LinkTo @route='login' class='nav-item d-none d-md-flex mr-2'>
-        <button class='btn btn-primary d-inline nav-link pl-2 pr-2' type='button'>
+        <button
```
wilco375, line 44: Or this

app/components/header-nav.hbs
```diff
@@ -41,8 +53,12 @@
   </div>
 
   {{#unless session.isAuthenticated}}
-    <button class='btn btn-locale btn-inverse-locale-{{intl.locale}}' title='{{t "component.headerNav.changeLocale"}}'
-      {{action 'toggleLocale'}} type='button'>
+    <button
```
wilco375, line 56: Or this

Ellen-Wittingen:
I cannot find a way to dynamically change the primary colors on the website. This is due to the fact that bootstrap colors are set in scss, and scss is compiled to css BEFORE the website is rendered, and is typically not rerendered afterwards. 
My other solution was to assign `$brand-primary: var(--primary-color)` in the varaibles.scss file, so it is prerendered with a css variable, which can be changed dynamically. However, bootstrap does not accept css variables :(

Only thing I came across is this, where they do some webpacker config magic to allow scss variable to be accessed in javascript but I do not know if it also allows reassinging them [https://css-tricks.com/getting-javascript-to-talk-to-css-and-sass/](https://css-tricks.com/getting-javascript-to-talk-to-css-and-sass/)

I also saw that bootstrap will suport themes in 5.3.0 which is probably perfect for our needs, but bootstrap's lastest release is still on 5.2.x and I dont see it updating this week. [https://getbootstrap.com/docs/5.3/customize/color-modes/](https://getbootstrap.com/docs/5.3/customize/color-modes/)

app/components/header-nav.hbs
```diff
@@ -1,5 +1,12 @@
...
-  <span class='left-sidebar-toggle d-block d-md-none link' {{action 'toggleLeftSidebar'}}>
+<nav
+  class='navbar container-fluid pb-0 pt-0 d-flex justify-content-between
+    {{if session.isAuthenticated "navbar-authenticated"}}'
+  aria-label='breadcrumb'
+>
+  <span
+    class='left-sidebar-toggle d-block d-md-none link'
+    {{action 'toggleLeftSidebar'}}
+  >
```
DrumsnChocolate, line 9: I don't like how this has been formatted still but it's getting reverted in two weeks anyway

PR #770, opened 2023-05-22
Remove quaestor check for copy usernames

wilco375:
### Summary
Turns out this feature is not only useful to the Quaestor but also for committee members (see email). This PR removes the check if the user can create debit collections (effectively if user is Quaestor).


DrumsnChocolate (approved):
I changed the base branch to staging. Did you intentionally select master?

PR #806, opened 2023-09-11
Fix sidebar rendering (for old-members)

DrumsnChocolate:
Turns out this broke in #805  . Reproduced locally, the problem only occurs right after login, any refresh already worked around the issue.

### DIFF
--- a/app/routes/profile.js
+++ b/app/routes/profile.js
@@ -5,7 +5,7 @@ export default class ProfileRoute extends AuthenticatedRoute {
   @service router;
 
   canAccess() {
-    return this.abilities.can('show individual users');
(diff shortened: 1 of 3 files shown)


app/routes/profile.js
```diff
@@ -5,7 +5,7 @@ export default class ProfileRoute extends AuthenticatedRoute {
   @service router;
 
   canAccess() {
-    return this.abilities.can('show individual users');
+    return this.abilities.can('show own users');
```
DrumsnChocolate, line 8: changing this is not strictly necessary, but it makes sense semantically to change it here too

wilco375 (approved):

PR #814, opened 2023-10-02
Removed quickpost from authenticated index page and moved bibleverse up

Ellen-Wittingen:
Removed the quickpost section from the authenticated index page, instead moved bible verse, photos, and polls in that space. On mobile version is the quickpost button also gone, and was bible verse moved to top as well.

A separate pull request will be opened that removes all quickpost functionality later.


DrumsnChocolate:
screenshots would make for easier review, but I'll check it out locally

DrumsnChocolate (approved):

PR #832, opened 2024-03-12
Update banner

ToMaarton:
Updated banner, also added correct links


app/components/advertisement-tool.js
```diff
@@ -31,100 +31,90 @@ export default Component.extend({
...
+      link: 'https://www.boeketcadeau.nl/',
+      image: 'advertisements/Boeketcadeau.png',
     },
     {
-      title: 'Led24',
-      link: 'https://www.led24.nl/led-verlichting/led-tl-buis/',
-      image: 'advertisements/led24.png',
+      title: 'Baktotaal',
+      link: 'https://www.baktotaal.nl/',
+      image: 'advertisements/Bouwhuis.png',
```
Ellen-Wittingen, line 46: Filename of image is "Bouwhuis" whilst the title is "Baktotaal". Its technically correct because they both represent the same website, but a but confusing. I recommend renaming the image file to "Baktotaal.png"

app/components/advertisement-tool.js
```diff
@@ -31,100 +31,90 @@ export default Component.extend({
...
     },
     {
-      title: 'Noppies',
-      link: 'https://www.noppies.com/',
-      image: 'advertisements/noppies.png',
+      title: 'Comfort Producten',
+      link: 'https://www.comfort-producten.nl/nl/',
+      image: 'advertisements/Comfort_products.png',
```
Ellen-Wittingen, line 51: Image name is inconsistent with the rest. Most images are camelCase (write words together but use a capital letter at the start of each word), this one is snake_case (using underscores to separate words). Since most are camelCase, please rename convert all image names to camelCase (should be about 5 images or so).

app/styles/components/advertisement-tool.scss
```diff
@@ -1,5 +1,5 @@
 // CSS3 code for animation of advertisements
-$amount-of-advertisements: 14;
+$amount-of-advertisements: 19;
```
Ellen-Wittingen, line 2: If you replace this with 18, the animation where it restarts at the first ad is smoother
```suggestion
$amount-of-advertisements: 18;
```

Ellen-Wittingen (changes requested):
Overall, things look good. Just have some minor comments. :)

app/components/advertisement-tool.js
ToMaarton: I made sure that all the image filenames are in the same convention I actually converted all of them to camelCase.

Ellen-Wittingen (approved):

PR #834, opened 2024-03-18
Fix/continuous delivery

Ellen-Wittingen:
If the workflow fails with an error message about blobs along the lines of the following: `buildx failed with: ERROR: failed to solve: [...] blob [...]` then all recent cache files for staging on the [Caches page](https://github.com/csvalpha/amber-ui/actions/caches) need to be deleted and the workflow needs to be executed again. Instructions for this were added to the README.md. 

Also added the option `Perform all jobs, regardless of whether there are actual changes?` to force the workflow to re-execute all its jobs. This was added because if the previous workflow run errored after staging was merged into master it will detect no changes and skip the Continuous Integration and Publish Image jobs, which means the changes won't go live.


.github/workflows/continuous-delivery.yml
```diff
@@ -7,6 +7,11 @@ on:
         description: Merge staging into master first? (y/N)
         required: false
         default: 'n'
```
guidojw, line 9: Can you make this one boolean too? I think I made it like this because boolean was not an option back in the day, but I see the UI nicely supports it now. Don't forget the checks below :)

guidojw (changes requested):

README.md
```diff
@@ -63,6 +63,8 @@ Deployments are done using GitHub Actions. To deploy a branch, follow the follow
 - Choose a branch and if you want to merge the changes on the staging branch into the master branch (only possible when the branch chosen in previous step is master).
 - Click the green button "Run workflow".
 
+*Note: If the workflow fails with an error message about blobs along the lines of the following: `buildx failed with: ERROR: failed to solve: [...] blob [...]`: remove all recent cache files for staging on the [Caches page](https://github.com/csvalpha/amber-ui/actions/caches) and try again. You might have to turn on the option `Perform all jobs, regardless of whether there are actual changes?` to force the workflow to re-execute all its jobs.*
```
guidojw, line 66: Odd bug, could this be a one-time bug caused by https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/? That is artifacts and not cache though, but it might be related.

DrumsnChocolate:
@lodewiges there's a problem with your commit: you have not taken care of the places where the value of this input is used. You can't change a type without updating the usages of the variable.
Will you be able to identify the occurrences in the workflow, and fix them? You can find all of these in `continuous-delivery.yaml`. An example is line 49:
```          INPUT_MERGE: ${{ github.event.inputs.merge }}```
the line is followed by usage of `INPUT_MERGE` where the variable is checked to be `'y'` or `'n'`. That will most certainly not result in expected behaviour now that we've changed it to a boolean.

**Review by DrumsnChocolate — CHANGES_REQUESTED** [review]
see comment

File: .github/workflows/continuous-delivery.yml [hunk 3]
```diff
@@ -61,7 +67,7 @@ jobs:
     name: Merge
     runs-on: ubuntu-latest
     needs: metadata
-    if: github.event.inputs.merge == 'y'
+    if: github.event.inputs.merge == true
```
DrumsnChocolate, line 70:
```suggestion
    if: github.event.inputs.merge
```
can't we do this?

.github/workflows/continuous-delivery.yml
```diff
@@ -43,7 +49,7 @@ jobs:
           INPUT_MERGE: ${{ github.event.inputs.merge }}
         run: |
           if [ "$GITHUB_REF_NAME" = 'master' ]; then
-            if [ "${INPUT_MERGE,,}" = 'y' ]; then
+            if [ "$INPUT_MERGE" = true ]; then
```
DrumsnChocolate, line 52: yes, surprisingly, this is how to evaluate boolean variables in shell... https://stackoverflow.com/a/2953673

DrumsnChocolate (approved):
only one last request

README.md
DrumsnChocolate: resolving because I don't know what to do with this and the remark does not seem prohibitive of merging this PR

.github/workflows/continuous-delivery.yml
lodewiges: What i found online is that ==true is better. source: https://stackoverflow.com/questions/2953646/how-can-i-declare-and-use-boolean-variables-in-a-shell-script/21210966#21210966

.github/workflows/continuous-delivery.yml
lodewiges: We were looking at the same site :)

.github/workflows/continuous-delivery.yml
lodewiges: we need to decide what would be better, and implement that everywhere, i forgot it in one line i saw.

DrumsnChocolate: You are talking about shell expressions, I'm talking about GitHub expressions: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions

They are separate concepts.

I've doublechecked why I suggested this. I come to a slightly more nuanced conclusion now, based on https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs
I think the conclusion is we can use `inputs.merge` and we don't need to check against a string, but when we use `github.event.inputs.merge` we do need to check against a string, because:
> The workflow will also receive the inputs in the github.event.inputs context. The information in the inputs context and github.event.inputs context is identical except that the inputs context preserves Boolean values as Booleans instead of converting them to strings. The choice type resolves to a string and is a single selectable option.

lodewiges: I think implemented it correctly can you check?

lodewiges:
bump

PR #876, opened 2024-10-10
Fix recent photo comments

### LINKED ISSUE #874: photo tags interfere with most recent comments
It looks like adding a photo tag makes the photo go to the most recently edited photo. in the past there where only comments that did this. I don't think it is relevant to show an old picture that has new tags instead showing new comments is more interesting

Is this intended behavior?

Example:
order I first placed 3 comments from test 1 to test 3 after this I added a tag to the second photo this made it go to the first place. 
Wanted behavior: the photo stays on the second place, this would mean that tags don't interfere with the order photo are shown
image

image

wilco375:
### Summary

Fixes #874

### DIFF
--- a/app/components/tools/recent-photos.hbs
+++ b/app/components/tools/recent-photos.hbs
@@ -7,13 +7,13 @@
   {{#if (can 'show photo-comments')}}
     <div class='card-body'>
       <div class='d-flex'>
-        {{#each photosWithComments as |photo|}}
+        {{#each latestComments as |comment|}}
           <LinkTo
             @route='photo-albums.photo-album.photos.photo'
-            @models={{array photo.photoAlbum.id photo.id}}
+            @models={{array comment.photo.photoAlbum.id comment.photo.id}}
             class='picture-link col-xs'
           >
-            <img class='featured-picture' src={{photo.imageThumbUrl}} />
+            <img class='featured-picture' src={{comment.photo.imageThumbUrl}} />
           </LinkTo>
         {{else}}
           <div class='p-2'>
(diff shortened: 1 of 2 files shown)


lodewiges:
when commenting twice on a photo it shows up twice in recent images. This is unwanted behavior.
image

image

lodewiges (changes requested):
when commenting twice on a photo it shows up twice in recent images. This is unwanted behavior.
image

image

wilco375:
> when commenting twice on a photo it shows up twice in recent images. This is unwanted behavior. image
> 
> image

Ah that's a good point. This might require some API changes to properly fix then. I'll look into it later!

DrumsnChocolate (approved):
what an interesting implementation oversight, haha
I trust the resulting implementation after the fix

DrumsnChocolate:
@wilco375 I don't think you need api changes; is it not possible to compute a list of photos and then run an operation of uniqueness over it? like a set or something like that

wilco375:
> @wilco375 I don't think you need api changes; is it not possible to compute a list of photos and then run an operation of uniqueness over it? like a set or something like that
Not if we want to get a fixed number of photos. Or we'd have to recursively get more photos from the backend until we have enough non-duplicate photos. But that would be nicer and more efficient to solve in a database query.

Either we get the photos directly and sort on date of latest comment => needs API addition of sorting by latest comment
Or we get latest comments but max one comment per picture => needs API addition of custom scope to filter out duplicate images
Or we repeatedly pull images from the API and do the filtering client side as described above, which might be very inefficient, especially if one image has a lot of recent comments (something that might not actually occur in practice but still)

DrumsnChocolate:
Right. I suppose the `sort: '-updated_at',` is passed to the API somehow? If that's the case, I suppose we could do some `sort: -commented_at`, if we can define some `commented_at` property in the backend. You're the expert on rails

DrumsnChocolate:
@wilco375 bump

DrumsnChocolate:
bump again :wink:

PR #877, opened 2024-10-10
Fix tag requires double click after first tag

### LINKED ISSUE #862: Tagging user in photo requires two clicks
To tag a user in a photo, two clicks are needed: one to open the popup, and one to focus the select/input box of the popup. Opening the popup should autofocus the select/input box.

wilco375:
### Summary

Fixes #862


lodewiges:
tested and works

DrumsnChocolate (approved):
I feel like this is also a sign of something not properly tearing down, but maybe that's just me? I would for some reason have expected selectApi to have been cleared. But I suppose the PhotoTags instance never really 'disappears' to Ember?

PR #884, opened 2024-10-11
Feat/codecov

wilco375:
### Summary
See #817 and #883


lodewiges (approved):
This should work now with the secrets token passed on in the continuous-delivery.yml

PR #887, opened 2024-10-14
album tag counter

ToMaarton:
(Issue#870) A small tag is added to the photo album cover cards, the number of photos that have been tagged out of the total amount of photos.

image

The tag does not show up for albums in which none of the photos have been tagged or that do not contain any photos.

The goal is to provide an easy way to check if the photos in the album have been tagged.

(Issue#863) The similar issue is fixed by also displaying the tag count of each photo in the same  box as the comment count.

image

### DIFF
--- a/app/components/cards/photo-album-card.hbs
+++ b/app/components/cards/photo-album-card.hbs
@@ -12,10 +12,10 @@
         <h3 class='card-subtitle'>{{moment-format album.date 'DD-MM-YYYY'}}</h3>
       </div>
       {{#if album.amountOfTags}}
-          <span class='album-tag-counter badge bg-info bottom-0 end-0'>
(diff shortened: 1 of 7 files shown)


app/components/cards/photo-album-card.hbs
```diff
@@ -5,11 +5,18 @@
       src='{{album.albumMediumUrl}}'
       class='card-img-rounded'
     />
+
```
lodewiges, line 8: i think this blank space shouldn't be here

lodewiges (approved):
code looks good, and works has some small reactivity issue but those should be solved by updating ember

wilco375:
@ToMaarton be careful with posting pictures containing personal information / photos of people. For example the gala photo in your last screenshot is not a public picture.

PR #889, opened 2024-10-26
removed quickposts and changed app routes

lodewiges:
Solves  #890 and #886 

remove all files regarding quick posts and changed some application routes to authentication routes because the should only be accessible when logged in.


app/router.js
```diff
@@ -203,17 +203,6 @@ Router.map(function () {
...
-  this.route('quickpost');
-
-  this.route('room-adverts', function () {
-    this.route('new');
-
-    this.route('room-advert', { path: ':id' }, function () {
-      this.route('edit');
-      this.route('destroy');
-    });
-  });
```
DrumsnChocolate: Not a _huge_ fan of removing the routes for the room-adverts we just added. I suspect this may have gone wrong during a merge? or you accidentally removed this. 
Either way, I've added it back.

DrumsnChocolate (approved):

app/router.js
lodewiges: This probably went wrong during a merge I did not mean to do this

DrumsnChocolate (changes requested):
hold up, we need to provide `canAccess` on AuthenticatedRoutes. Did you test the pages you changed the route classes for?

lodewiges:
> hold up, we need to provide `canAccess` on AuthenticatedRoutes. Did you test the pages you changed the route classes for?

I checked a couple of pages, I did not do I did every single one because the changes were the same between pages.

DrumsnChocolate:
the pages did break; with canAccess the authenticatedRoute does not want to function properly.

Let's look at `application.js` and see the implementation of the `AuthenticatedRoute`: this class is a bit confusing, in that it fulfils two responsibilities: it does not only check for _authentication_ but also verifies _authorization_. I find these are very similar terms, so let's be clear:
- Authentication is the process of verifying authenticity. -> Login
- Authorization is the process of verifying authority. -> Permissions

These are two different concepts. I think that we may benefit from separating these responsibilities, because there are definitely cases where we can use a logged-in check but do not need a permissions check.
I propose the following classes:
`ApplicationRoute` as it is now
`AuthenticatedRoute`:
- extends `ApplicationRoute`
- verifies authentication

`AuthorizedRoute`, 
- extends `AuthenticatedRoute`
- verifies authority

DrumsnChocolate:
In the meantime I'll add a `canAccess` to all the relevant routes.
We also don't need to introduce new classes in this PR, we should probably make an issue for my comment above

DrumsnChocolate (approved):
afbeelding

DrumsnChocolate:
@lodewiges did you also open a related PR on amber-api? I don't think it makes sense to keep the models and endpoints when all the frontend is gone

lodewiges:
> @lodewiges did you also open a related PR on amber-api? I don't think it makes sense to keep the models and endpoints when all the frontend is gone

no not yet i first wanted the frontend to be gone before doing that.

Issue #845, opened 2024-05-29
Add "Tweede kamer" link to scheduling document


DrumsnChocolate:
Should not be too hard to add this next to this button 
![afbeelding](https://github.com/csvalpha/amber-ui/assets/32799956/6616b17e-825d-4bef-9c35-93845ac6fe5f)

lodewiges:
what should the different statuses be? for now i named them; Chilling, Studying and Banaan. i could imagine have one for being in a meeting

DrumsnChocolate:
This issue doesn't describe a request for some kind of status. What are you referring to?

This issue is asking for: a button pointing to a nextcloud spreadsheet.

How did you interpret this issue/how are you solving it?

lodewiges:
Okay interpreted it wrong, i thought the idea was to have a second button were people could show that they were studying, chilling or having a meeting in the 2de kamer.

i am not sure how i misinterperet this, most likly combination of tunnelvision and dyslexia.

PR #635, opened 2022-10-23
Refactor: `{{action}}` helper usage

DrumsnChocolate:
https://guides.emberjs.com/release/upgrading/current-edition/action-on-and-fn/
base branch will be updated :sparkles: automatically :sparkles: when #556 merges into staging

ref #899 

this is not the whole thing, I am opening a sequel PR soon. For the sake of keeping PRs reviewable.

How to review this:
- [ ] have a look at some of the changes, to get an idea of what this PR is doing.
- [ ] manually test all pages that have been affected by this PR, and see if the buttons on those pages work as expected.

PR #891, opened 2024-10-27
Implementing a button for Presence in the 2de kamer

lodewiges:
adds #845

There were 2 ways I was considering implementing it, making a new component or rewriting the original component so that both presences indicators could be taken care of with one controller. I decided to implement a second controller because there are multiple difference between the 2 presences indicators.

needs https://github.com/csvalpha/amber-api/pull/449


DrumsnChocolate:
cool! I will review this soon. Either this weekend or at our meeting next week

app/components/tools/study-room-presence.js
DrumsnChocolate, line 1: this component is written in the old style, please follow the new octane component style. There are a lot of examples that use glimmer components already

specifically, you can take example from https://github.com/csvalpha/amber-ui/pull/635/files#diff-b264f9e9a3db1a61c4d3c4c38a3c454a93a33286c5d2b1e820513bc9d4986084 where I've rewritten the board-room-presence component

app/components/tools/study-room-presence.hbs
DrumsnChocolate, line 1: please also write this template in octane style. Some quick rules (without regard for whether you followed these or not):
- use angle bracket syntax for inserting components
- refer to controller properties/functions with `this.` i.e. `this.property` or `this.someFunction`
- refer to arguments passed to this component using `@`. So, if the parent component has passed an argument called `someArgument`, then we can refer to it in this template with `@someArgument`.
- do not use the action helper anymore. see #635 and #896 for examples of what to write instead.

please thoroughly familiarize yourself with #429 and the tutorials/guides/resources mentioned there.

DrumsnChocolate (changes requested):
Looking very good overall! Differences between this and board-room-presence are abstractable, but that doesn't mean we should. Often, it is better to not introduce abstraction when it does not serve a practical purpose other than reducing the amount of code. This code is more maintainable as it is, because abstraction would introduce complexity.

I have left some remarks, let me know when you've implemented them.

app/components/tools/study-room-presence.hbs
```diff
@@ -0,0 +1,137 @@
+<div class='room-presence-dropdown page-action'>
+  <BsButton @type='primary'>
+    <FaIcon @icon='book' />
+    <span class='d-none d-md-inline'>2de kamer</span>
+    <span class='status-dot label-{{this.overallStatus}}'></span>
+
+    <BsPopover
+      {{!-- @triggerEvents='focus' --}}
```
DrumsnChocolate, line 8: why is there a comment?

app/components/tools/study-room-presence.js
```diff
@@ -0,0 +1,128 @@
+import { dropTask, restartableTask, timeout } from 'ember-concurrency';
+import Component from '@glimmer/component';
+// import Ember from 'ember';
```
DrumsnChocolate, line 3:
```suggestion
```

DrumsnChocolate (approved):

app/components/tools/study-room-presence.hbs
DrumsnChocolate, line 8:
```suggestion
```

PR #804, opened 2023-09-09
Feature: Room adverts

Ellen-Wittingen:
Added room adverts. They are accessible via [http://localhost:4200/room-adverts](http://localhost:4200/room-adverts) and via the menu-item in the left sidebar.

There is also a view for unauthorized users on the public part of the website, it is accessible via [http://localhost:4200/public/room-forum](http://localhost:4200/public/room-forum) and via the menu-item "Kamerforum" in the "Over ons" dropdown.

Please also check if the English translations are proper English, as nobody except from me has looked at them so far.

Note: it requires the following amber-api branch to function properly: [https://github.com/csvalpha/amber-api/pull/393](https://github.com/csvalpha/amber-api/pull/393) NOTE BY MATTEO: this backend branch has already been incorporated into staging

PR #894, opened 2024-11-03
Fix: sidebar toggle on mobile

DrumsnChocolate:
self-descriptive. probably broke in #804


lodewiges (approved):
looks good unable to test

PR #896, opened 2024-11-03
Refactor: `{{action}}` helper usage, the sequel

DrumsnChocolate:
## DO NOT MERGE BEFORE #635  IS MERGED
bases on #635
ref #899 
may require a third PR after this

How to review this:
- [ ] have a look at some of the changes, to get an idea of what this PR is doing.
- [ ] manually test all pages that have been affected by this PR, and see if the buttons on those pages work as expected.

## New prevent-default helper
This PR also introduces a new template helper: `prevent-default`. It is used to wrap the action we assign to be executed when an event takes place. Specifically, it adds a call to `event.preventDefault()`. Why would we do this in a helper? Well:
1. because we don't want to pollute the js files with `e.preventDefault();` calls. 
2. the decision of whether or not to call `preventDefault` should take place in the template, because that's where it is easiest to determine.

To elaborate on point two, when we have the following action:
```js
@action
someAction() {
    someLogic();
}
```
Then a template with 
```hbs
<form {{on 'submit' someAction}}></form>
``` 
requires a call to `e.preventDefault()`:
```js
@action
someAction(event) {
    event.preventDefault();
    someLogic();
}
```
otherwise the page is reloaded automatically (because that's what a submit event on a form does, apparently.) Now, suppose instead the template contains 
```hbs
<ComponentForDoingSomething @doThis={{someAction}}></ComponentForDoingSomething>
```
then we _don't_ need a `e.preventDefault()` call... See how that's confusing? Sidenote: in reality the underlying ComponentForDoingSomething generally still performs a call to `e.preventDefault`, but that's none of our business, since principles like [dependency inversion](https://en.wikipedia.org/wiki/Dependency_inversion_principle) dictate that we do not concern ourselves with the implementation details of a child component, but only with the parts we can 'see' of that child component.

Now let's make the point even better: if we have a problematic template that contains **both** of these types of usages: 
```hbs
<form {{on 'submit' someAction}}></form>
<ComponentForDoingSomething @doThis={{someAction}}></ComponentForDoingSomething>
```
 then what do we do? We need to call `e.preventDefault()` for one element in the template, but not for the other. But that's easily solved with the new helper:
 ```hbs
 <form {{on 'submit' (prevent-default someAction)}}></form>
 <ComponentForDoingSomething @doThis={{someAction}}></ComponentForDoingSomething>
```


DrumsnChocolate:
I initially began rewriting components to glimmer when I encountered them, but that's quite a lot of work and not strictly related to the task this PR is really meant for. I've found it more efficient to just refactor all action helpers and leave other non-octane things as they are for now. We can tackle them in other PRs.

I've also not tested that every action occurrence I've refactored in in this PR still functions properly. Not yet at least. I did test that for #635 , but I figured it's more efficient to first rewrite, and then go through all the rewrites and test their functioning.

DrumsnChocolate:
also fixes some styling in the print-enrolled component.
before:
afbeelding

after:
afbeelding
not perfect but quick win

lodewiges (changes requested):
LGTM 

i think you forgot 4 files
app/components/advertisement-tool.js
app/components/boolean-tag.js
app/components/cards/activity-card.js
app/components/cards/photo-album-card.js
when these are updated i will aprove

lodewiges (approved):

PR #913, opened 2024-11-21
Fix/sofialink

lodewiges:
edited the link in envoriment so the staging site links to stagingstreep.csvalpha.nl instead of streep.csvalpha.nl

### DIFF
--- a/app/components/menu-sidebar.js
+++ b/app/components/menu-sidebar.js
@@ -10,16 +10,6 @@ export default class MenuSidebar extends Component {
   @service store;
   @tracked staticPages;
 
-  get sofiaUrl() {
-    if (config.environment === 'development') {
-      return 'http://localhost:5000';
-    } else if (config.deployTarget === 'staging') {
-      return 'https://stagingstreep.csvalpha.nl';
-    } else {
-      return 'https://streep.csvalpha.nl';
-    }
-  }
-
   @action
   invalidateSession() {
     this.session.invalidate();
--- a/app/components/menu-sidebar.js
+++ b/app/components/menu-sidebar.js
@@ -28,4 +28,8 @@ export default class MenuSidebar extends Component {
       this.staticPages = staticPages;
     });
   }
+  
+  get sofiaUrl() {
+    return config.sofiaUrl;  // Access the sofiaUrl from config
+  }
 }
--- a/app/components/menu-sidebar.js
+++ b/app/components/menu-sidebar.js
@@ -28,8 +28,8 @@ export default class MenuSidebar extends Component {
       this.staticPages = staticPages;
     });
   }
-  
+
   get sofiaUrl() {
-    return config.sofiaUrl;  // Access the sofiaUrl from config
+    return config.sofiaUrl; // Access the sofiaUrl from config
   }
 }
--- a/app/components/menu-sidebar.js
+++ b/app/components/menu-sidebar.js
@@ -13,7 +13,7 @@ export default class MenuSidebar extends Component {
   get sofiaUrl() {
     if (config.environment === 'development') {
       return 'http://localhost:5000';
-    } else if (config.environment === 'staging') {
+    } else if (config.deployTarget === 'staging') {
       return 'https://stagingstreep.csvalpha.nl';
     } else {
       return 'https://streep.csvalpha.nl';
--- a/app/components/menu-sidebar.js
+++ b/app/components/menu-sidebar.js
@@ -13,7 +13,7 @@ export default class MenuSidebar extends Component {
   get sofiaUrl() {
     if (config.environment === 'development') {
       return 'http://localhost:5000';
-    } else if (config.deployTarget === 'staging') {
+    } else if (config.environment === 'staging') {
       return 'https://stagingstreep.csvalpha.nl';
     } else {
       return 'https://streep.csvalpha.nl';
--- a/app/components/menu-sidebar.hbs
+++ b/app/components/menu-sidebar.hbs
@@ -80,7 +80,7 @@
         />
 
         {{! Link to SOFIA }}
-        <a class='link' href={{this.sofiaUrl}} target='_blank'>
+        <a class='link' href={{config.sofiaUrl}} target='_blank'>
           <MenuSidebarItem
             @link=''
             @hasPermission={{true}}
--- a/app/components/menu-sidebar.hbs
+++ b/app/components/menu-sidebar.hbs
@@ -80,7 +80,7 @@
         />
 
         {{! Link to SOFIA }}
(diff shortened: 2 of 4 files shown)


app/components/menu-sidebar.js
```diff
@@ -13,7 +13,7 @@ export default class MenuSidebar extends Component {
   get sofiaUrl() {
     if (config.environment === 'development') {
       return 'http://localhost:5000';
-    } else if (config.environment === 'staging') {
+    } else if (config.deployTarget === 'staging') {
```
DrumsnChocolate, line 16: we are now using two variables here, is that wise?

lodewiges: No, that is not the ideal situation. but is better then the situation now

lodewiges: okay, i did some more research it is possible to use one variable however we need to rewrite logic in the anvirment.js file and change the dockerfiles on the server

app/components/menu-sidebar.js
```diff
@@ -13,7 +13,7 @@ export default class MenuSidebar extends Component {
   get sofiaUrl() {
     if (config.environment === 'development') {
```
lodewiges, line 14:
```suggestion
   if (config.deployTarget === 'production') {
      return 'https://streep.csvalpha.nl';
    } else if (config.deployTarget === 'staging') {
      return 'https://stagingstreep.csvalpha.nl';
    } else {
      return 'http://localhost:5000';
    }
```

lodewiges: This could also be a option however it does change default behaviour.

app/components/menu-sidebar.js
DrumsnChocolate: Okay so my point is that `environment` and `deployTarget` seem to be used in similar context in similar ways. Why do we distinguish between them? Can you explain?

There are currently four places where `config.environment` is used, and I don't see why we couldn't use `config.deployTarget` instead. How are these two variables really different? 
afbeelding

lodewiges: I didn't write the original code, so I can't explain it with certainty. I think the general idea was that we have three environments: test, development, and production. All three are different environments and sometimes require different dependencies.

within the production environment we have 2 deploy targets: production and staging, it is confusing that we use production twice. the difference between these deploy targets are configurations they should use the same dependencies.

This leads me to think the second suggestion is the better of the 2

DrumsnChocolate: okay that does clear things up a bit, I'm still confused about some stuff but I'll try to formulate my thoughts:
- My intuition tells me that direct access to `deployTarget` is not something the `menu-sidebar.js` needs to be responsible for. It should not have to think too much about what the sofiaUrl is, just like `router.js` does not have to think too hard about what `rootUrl` is. So, my recommendation is to move the logic for determining the sofiaUrl out of `menu-sidebar.js` and into `environment.js`.
- I agree that using `production` twice is confusing. I'm not sure what the `production` environment refers to, but the other two should be clear. I think that the `production` environment might really mean a `real-world` environment? There's a bunch of words for it, `build` could also be an alternative. Then you would have `test`, `development`, `build`.
- You say there are **2** deploy targets, I would say there is an implicit deployTarget: `local`. You can recognize this in `environment.js` by seeing that `ENV.clientId` has a fall back value for when `deployTarget` is neither `production` or `staging`. We don't _need_ to write this down because it requires extra work 
- I feel like the `sofiaUrl` should not depend on `environment`, but on `deployTarget`. In general, I think most config values should not depend on what environment we're running _in_, but what target we're running _on_. The `environment` feels more like a 'mode' of running, while `deployTarget` feels like a 'place'.
However, there are still things we definitely should turn off or on based on which 'mode' we're in. Two examples are **sentry** and apparently **service workers** (no clue what those are).
Like you mentioned, different environments may require different dependencies.


## TLDR
- I identify three envs: `test`, `development`, `production` (or `build`, we could rename to that)
- I identify three deploy targets: `local`, `staging`, `production`
- we should define `config.sofiaUrl`, and determine its value in `environment.js` where `config.rootUrl` is also defined.
- we should stop using `config.environment` to determine `config.sofiaUrl`: it should only really depend on `config.deployTarget` i.e. _where_ is the code running.

DrumsnChocolate: (yes I'm aware that we're going beyond the scope of this PR, I hope you don't mind. if you prefer to get staging working first, and discuss details like this later, I can approve this PR and you can make my suggested changes in another PR)

lodewiges: Okay, I agree with the first point you are making. 

 As far I understand this is correct "I think that the production environment might really mean a real-world environment" Build would be a better alternative. I can make an issue for this. 

However I don't agree with the third point, there are 2 deploy targets we deploy to using GitHub actions. We dont deploy to local or test but we do need a variable for clientID otherwise we are unable to test/connect to amber-api. 

I have no problem with discussing it thoroughly, I learn alot doing this so i am happy.

DrumsnChocolate: Note before reading: I have not checked the latest commits on this, simply replying to this discussion.

Oke let's phrase it this way: the `deployTarget` should have three possible values: `'staging'` `'production'`, and `undefined` i.e. we're not deploying anywhere. Whenever you are running locally or maybe running end to end tests on ci, this third value comes into play. So you still have to take this third option into account whenever you are accessing `deployTarget` in any logic. Even though `undefined` it is not formally a deploy target that you can physically point to in the real world, it signifies something about where we deploy (or rather, that we are not deploying) and it impacts what the URLs should be.

But **more imporantly**, aside from our disagreement around the idea that there are three 'deploy targets', do you agree with the points I made in the TLDR of the comment? That is:
- we should define `config.sofiaUrl`, and determine its value in `environment.js` where `config.rootUrl` is also defined.
- we should stop using `config.environment` to determine `config.sofiaUrl`: the latter should only depend on `config.deployTarget`.

lodewiges: I agree with you on both point, is it good as is or do we want to move.
'''
ENV.clientId = '123456789';
ENV.sofiaurl = 'http://localhost:5000';
'''
to the general section or let them stay where they are

DrumsnChocolate (approved):

app/components/menu-sidebar.js
DrumsnChocolate: > I agree with you on both point, is it good as is or do we want to move.
> 
> ```
> ENV.clientId = '123456789';
> ENV.sofiaurl = 'http://localhost:5000';
> ```
> 
> to the general section or let them stay where they are

I don't have a strong opinion on this

PR #939, opened 2025-02-17
Remove IBAN option from group export

lodewiges:
Fixes https://github.com/csvalpha/amber-ui/issues/940

I removed the ability to download all IBAN because it is not stored in the user table but in the mandates table. besides that it is really sensitive data.

If we need this functionality we need to a export button to the mandates tab or fix it some other way


DrumsnChocolate (approved):
Are we sure the avatar url is never used in the export?

Issue #888, opened 2024-10-21
Phototag counters are visible on external website


lodewiges:
![image](https://github.com/user-attachments/assets/c1eedaf5-8321-4a93-b938-923cc56bf125)
the phototag counter is visible on the external website, photo tags itself are not visible

lodewiges:
same problem with comment counters

PR #946, opened 2025-02-25
Fix for multiple bugs regarding photoalbums

lodewiges:
Fixes https://github.com/csvalpha/amber-ui/issues/888
Remove the ability to view reactions for external users due to it not adding a lot of value and make sure that no inappropriate content can be viewed publicly. 

DISADVANTAGE

When publicly viewing a image the screen looks empty. This could be improved by adding 2 arrows next to the picture instead of in the navbar. we could also consider having the picture in forground and the complete overview of the fotoalbum in the background

### DIFF
--- a/app/components/cards/photo-album-card.hbs
+++ b/app/components/cards/photo-album-card.hbs
@@ -11,7 +11,7 @@
         <h2 class='card-title'>{{album.title}}</h2>
         <h3 class='card-subtitle'>{{moment-format album.date 'DD-MM-YYYY'}}</h3>
       </div>
-      {{#if album.amountOfTaggedPhotos}}
+      {{#if (and album.amountOfTaggedPhotos (can 'show photo-tag'))}}
         <span class='album-tag-counter badge bg-info bottom-0 end-0'>
           <FaIcon @icon='tag' />
           {{album.amountOfTaggedPhotos}}/{{album.amountOfPhotos}}
--- a/app/templates/photo-albums/photo-album/photos/index.hbs
+++ b/app/templates/photo-albums/photo-album/photos/index.hbs
@@ -1,6 +1,6 @@
 <EmberWormhole @to='navbar-wormhole'>
   {{#page-actions-buttons pageActions}}
-    {{#if (can 'show users')}}
+    {{#if this.session.isAuthenticated}}
       <div class='col col-md-auto page-action'>
         <button class='btn btn-primary' {{action 'downloadAlbum'}} type='button'>
           <FaIcon @icon='download' />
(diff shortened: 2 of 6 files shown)


DrumsnChocolate:
> When not publicly viewing a image the screen looks empty

Could you share some screenshots of what things look like now?

app/components/cards/photo-album-card.hbs
```diff
@@ -11,7 +11,7 @@
         <h2 class='card-title'>{{album.title}}</h2>
         <h3 class='card-subtitle'>{{moment-format album.date 'DD-MM-YYYY'}}</h3>
       </div>
-      {{#if album.amountOfTaggedPhotos}}
+      {{#if (and album.amountOfTaggedPhotos (can 'show photo-tag'))}}
```
DrumsnChocolate, line 14: This could also be shut tight on the backend side; if "can show photo-tag" is not applicable, we could simply have album.amountoftaggedphotos be 0. Otherwise it seems to me that there's information leaking somewhere

app/templates/photo-albums/photo-album/photos/index.hbs
```diff
@@ -1,11 +1,13 @@
 <EmberWormhole @to='navbar-wormhole'>
   {{#page-actions-buttons pageActions}}
-    <div class='col col-md-auto page-action'>
-      <button class='btn btn-primary' {{action 'downloadAlbum'}} type='button'>
-        <FaIcon @icon='download' />
-        <span class='d-none d-md-inline'>Fotoalbum downloaden</span>
-      </button>
-    </div>
+    {{#if (can 'show users')}}
```
DrumsnChocolate, line 3: Why specifically this permission?

DrumsnChocolate (approved):

app/components/cards/photo-album-card.hbs
DrumsnChocolate: This remark holds for other similar properties that appear in this PR, too.

lodewiges:
> > When not publicly viewing a image the screen looks empty
> 
> Could you share some screenshots of what things look like now?

image

app/components/cards/photo-album-card.hbs
lodewiges: good call. however, we count photo tags clientside, so it is not yet relevant. i added to the issue which would count photo tags serverside

app/templates/photo-albums/photo-album/photos/index.hbs
lodewiges: i needed a way to check if somebody was logged in, i just now realized there is a better way

PR #956, opened 2025-03-09
make sidekiq UI modifiable & Be rijbewijs improvements

lodewiges:
Fixes https://github.com/csvalpha/amber-api/issues/524 & https://github.com/csvalpha/amber-api/issues/424
I also sorted the user model a bit better

requires https://github.com/csvalpha/amber-api/pull/528

### DIFF
--- a/app/models/user.js
+++ b/app/models/user.js
@@ -26,7 +26,10 @@ export default class User extends Model {
   // Preferences / settings
   @attr foodPreferences;
   @attr vegetarian;
+  @attr trailerDriversLicense;
   @attr('date-only') startStudy;
+
+  // Restricted Settings
   @attr emergencyContact;
   @attr emergencyNumber;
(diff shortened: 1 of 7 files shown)


app/models/user.js
```diff
@@ -51,6 +52,8 @@ export default class User extends Model {
   @attr('date') activatedAt;
   @attr('date') createdAt;
   @attr('date') updatedAt;
+  @attr sidekiqAccess;
+  @attr setupComplete;
```
DrumsnChocolate, line 56: Why do we consider this a security property?

DrumsnChocolate (approved):

app/models/user.js
lodewiges: both should be technical wrong copied them wrong

lodewiges: there considered a technical property

PR #971, opened 2025-04-14
Enable lustrum theme

lodewiges:
Enables lustrum theme, This PR should be pushed monday 9 o'clock 

did not check mobile so maybe pix should be converted to 2.5 rem


DrumsnChocolate:
Screenshot?

app/styles/dies-theme.scss
```diff
@@ -1,18 +1,18 @@
...
+  $app-bg-color: #c8ab8b;
+  $sidebar-fg-color: #c8ab8b;
+  $sidebar-bg-color: #3b4231;
+  $navbar-fg-color: #c8ab8b;
+  $navbar-bg-color: #3b4231;
+  $navbar-button-fg-color: #c8ab8b;
+  $navbar-button-bg-color: #3b4231;
   $dies-logo-url: '/images/dies_logo.png';
-  $dies-logo-width: 5rem;
+  $dies-logo-width: 40px;
```
DrumsnChocolate, line 15: Yeah please keep using rem

DrumsnChocolate (changes requested):

lodewiges:
> Screenshot?

image
backend is offline so not all elements are visible

app/styles/dies-theme.scss
lodewiges: yeah should prob be 2.5rem

PR #977, opened 2025-04-24
Fix for "Download fotos" button not showing while being authenticated

lodewiges:
One of the old members notified us about functionality not working properly and suggested this fix.


RubenSmit:
Lekker bezig!

PR #990, opened 2025-05-19
General improvement

lodewiges:
Solves https://github.com/csvalpha/amber-ui/issue/982 make menu button fully visible
Solves https://github.com/csvalpha/amber-ui/issue/911 change phototag centerlocation

In progress search button on mobile
TO DO 

- [x] make filter make entire length
- [x] make the 2 icons in the top right same size 
image


lodewiges:
there are still 2 things i would like to fix before going to prod 
image

- [ ] we should introduce a margin around the searchbar
- [ ] and we should increase the size of the burger menu to 1.4rem to match the other icons

